PR #5, opened 2016-06-29
Develop - fixes and extensions needed by large barrel data analysis

alekgajos:
- fixed two fields with confused meaning in the text database
- added a JPetAuxilliaryData class for data passed between modules
- fixed naming of intermediate and final files produced by the embedded HLD unpacker
- miscellaneous minor fixes

### DIFF
--- a/JPetAuxilliaryData/JPetAuxilliaryData.cpp
+++ b/JPetAuxilliaryData/JPetAuxilliaryData.cpp
@@ -1,0 +1,85 @@
+/**
+ *  @copyright Copyright 2016 The J-PET Framework Authors. All rights reserved.
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may find a copy of the License in the LICENCE file.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  @file JPetAuxilliaryData.cpp
+ */
+
+#include "JPetAuxilliaryData.h"
+#include "../JPetCommonTools/JPetCommonTools.h"
+#include "../JPetLoggerInclude.h"
+#include <sstream>
+#include <limits>
+
+ClassImp(JPetAuxilliaryData);
+
+const double JPetAuxilliaryData::Unset = -std::numeric_limits<double>::infinity();
+
+JPetAuxilliaryData::JPetAuxilliaryData():
+  TNamed("Auxilliary Data", "")
+{
+}
+
+JPetAuxilliaryData::JPetAuxilliaryData(std::string name):
+  TNamed(name.c_str(), name.c_str())
+{
+}
+
+void JPetAuxilliaryData::createVector(std::string name, int initial_size){
+  fVectors[name].resize(initial_size, Unset);
+}
+
+void JPetAuxilliaryData::createMap(std::string name){
+  fDictionaries[name];
+}
+
+void JPetAuxilliaryData::setValue(std::string container_name, unsigned int index, double value){
+  if( fVectors.count(container_name) > 0 &&
+      index < fVectors[container_name].size() ){
+    fVectors[container_name][index] = value;
+  }
+}
+
+void JPetAuxilliaryData::setValue(std::string container_name, std::string key, double value){
+  if( fDictionaries.count(container_name) > 0 ){
+    fDictionaries[container_name][key] = value;
+  }
+}
+
+double JPetAuxilliaryData::getValue(std::string container_name, unsigned int index) const {
+  if( fVectors.count(container_name) > 0 ){
+    if(index < fVectors.at(container_name).size() ){
+      return fVectors.at(container_name).at(index);
+    }else{
+      WARNING(Form("JPetAuxilliaryData: Index %d is out of bounds"
+		   " of the vector \"%s\"", index, container_name.c_str()));
+    }
+  }else{
+    WARNING(Form("JPetAuxilliaryData: Referring to a non-existent container \"%s\"",
+		 container_name.c_str()));
+  }
+  return JPetAuxilliaryData::Unset;
+}
+
+double JPetAuxilliaryData::getValue(std::string container_name, std::string key) const {
+  if( fDictionaries.count(container_name) > 0 ){
+    if( fDictionaries.at(container_name).count(key) > 0 ){
+      return fDictionaries.at(container_name).at(key);
+    }else{
+      WARNING(Form("JPetAuxilliaryData: Referring to a non-existent key \"%s\""
+		   " in a map \"%s\"", key.c_str(), container_name.c_str()));
+    }
+  }else{
+    WARNING(Form("JPetAuxilliaryData: Referring to a non-existent container \"%s\"",
+		 container_name.c_str()));
+  }
+  return JPetAuxilliaryData::Unset;
+}
--- a/JPetAuxilliaryData/JPetAuxilliaryData.cpp
+++ b/JPetAuxilliaryData/JPetAuxilliaryData.cpp
@@ -34,10 +34,18 @@ JPetAuxilliaryData::JPetAuxilliaryData(std::string name):
 }
 
 void JPetAuxilliaryData::createVector(std::string name, int initial_size){
+  if( fVectors.count(name) > 0 ){
+    WARNING("JPetAuxilliaryData: A vector with the requested name already exists and will be overwritten.");
+    fVectors[name].clear();
+  }
   fVectors[name].resize(initial_size, Unset);
 }
 
 void JPetAuxilliaryData::createMap(std::string name){
+  if( fDictionaries.count(name) > 0 ){
+    WARNING("JPetAuxilliaryData: A map with the requested name already exists and will be overwritten.");
+    fDictionaries[name].clear();
+  }
   fDictionaries[name];
 }
 
--- a/JPetHLDReader/JPetHLDReader.h
+++ b/JPetHLDReader/JPetHLDReader.h
@@ -75,6 +75,11 @@ public:
     else return false;
   }
 
+  virtual TObject* getObject(const char* name){
+    if (fFile) return fFile->Get(name);
+    else return 0;
+  }
+  
 protected:
   virtual bool openFile(const char* filename);
   virtual bool loadData(const char* treename = "T");
--- a/JPetHLDReader/JPetHLDReader.h
+++ b/JPetHLDReader/JPetHLDReader.h
@@ -75,7 +75,7 @@ public:
     else return false;
   }
 
-  virtual TObject* getObject(const char* name){
+  virtual TObject* getObjectFromFile(const char* name){
     if (fFile) return fFile->Get(name);
     else return 0;
   }
--- a/JPetLayer/JPetLayer.h
+++ b/JPetLayer/JPetLayer.h
@@ -19,7 +19,7 @@
 #include <TRef.h>
 #include "../JPetFrame/JPetFrame.h"
 #include "TNamed.h"
-
+#include <cassert>
 
 /**
  * @brief Parametric class representing database information on a single cyllindrical layer of a JPetFrame.
@@ -39,9 +39,9 @@ public:
   JPetLayer();
   JPetLayer(int id, bool isActive, std::string name, float radius);
   
-  inline bool operator==(const JPetLayer& layer) const { return getId() == layer.getId(); }
-  inline bool operator!=(const JPetLayer& layer) const { return getId() != layer.getId(); }
-  
+  bool operator==(const JPetLayer& layer) const;
+  bool operator!=(const JPetLayer& layer) const;
+
   inline int getId() const { return fId; }
   inline bool getIsActive() const { return fIsActive; }
   inline std::string getName() const { return fName; }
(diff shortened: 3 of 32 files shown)


JPetRawSignal/JPetRawSignal.h
```diff
@@ -88,29 +88,40 @@ class JPetRawSignal: public JPetBaseSignal
...
   inline int getNumberOfLeadingEdgePoints(void) const {
     return fLeadingPoints.size();
   }
   inline int getNumberOfTrailingEdgePoints(void) const {
     return fTrailingPoints.size();
   }
 
-  const JPetSigCh & getTOTPoint() const;
-  void setTOTPoint(const JPetSigCh & totSigCh);
+  /* const JPetSigCh & getTOTPoint() const; */
```
wkrzemien: Don't leave commented code.
Remove  it if not necessary.

JPetRawSignal/JPetRawSignal.h
```diff
@@ -88,29 +88,40 @@ class JPetRawSignal: public JPetBaseSignal
...
+  /* void setTOTPoint(const JPetSigCh & totSigCh); */
 
   /**
    * @brief Get the Time Over Threshold value in [ps]
    *
    * @return Time Over Threshold value in [ps] or JpetSigCh::kUnset if TOT was not registered for this signal.
    */
-  double getTOT() const;
+  /* double getTOT() const; */
```
wkrzemien: As above

JPetRawSignal/JPetRawSignalTest.cpp
```diff
@@ -9,20 +9,20 @@
 // Tests
 BOOST_AUTO_TEST_SUITE(ParamDataTS)
 
-BOOST_AUTO_TEST_CASE(ToTTest) {
-  JPetSigCh sigch;
-  sigch.setType(JPetSigCh::Charge);
-  sigch.setValue(17.f);
-  JPetRawSignal signal;
-  signal.addPoint(sigch);
+/* BOOST_AUTO_TEST_CASE(ToTTest) { */
```
wkrzemien: Remove commented code

JPetRawSignal/JPetRawSignalTest.cpp
```diff
@@ -9,20 +9,20 @@
...
+/*   JPetSigCh sigch; */
+/*   sigch.setType(JPetSigCh::Charge); */
+/*   sigch.setValue(17.f); */
+/*   JPetRawSignal signal; */
+/*   signal.addPoint(sigch); */
 
-  BOOST_REQUIRE_EQUAL(signal.getTOT(), 17.f);
+/*   BOOST_REQUIRE_EQUAL(signal.getTOT(), 17.f); */
```
wkrzemien: The same

JPetRawSignal/JPetRawSignalTest.cpp
```diff
@@ -9,20 +9,20 @@
...
 
-  JPetRawSignal signal2;
-  signal2.setTOTPoint(sigch);
+/*   JPetRawSignal signal2; */
```
wkrzemien: The same

JPetRawSignal/JPetRawSignalTest.cpp
```diff
@@ -9,20 +9,20 @@
...
+/*   signal2.setTOTPoint(sigch); */
 
-  BOOST_REQUIRE_EQUAL(signal2.getTOT(), 17.f);
-}
+/*   BOOST_REQUIRE_EQUAL(signal2.getTOT(), 17.f); */
```
wkrzemien: The same

JPetLayer/JPetLayer.h
```diff
@@ -39,8 +39,8 @@ class JPetLayer: public TNamed
   JPetLayer();
   JPetLayer(int id, bool isActive, std::string name, float radius);
   
-  inline bool operator==(const JPetLayer& layer) { return getId() == layer.getId(); }
-  inline bool operator!=(const JPetLayer& layer) { return getId() != layer.getId(); }
+  inline bool operator==(const JPetLayer& layer) const { return getId() == layer.getId(); }
```
wkrzemien: Comment concerning not const, but  the idea of this operator overloading, that I dont like in that shape.
Since the object contains not only id, but also e.g. radius, name etc it is possible to create two objects with the same id but different radius and == will say that it is the same object, even that one can check that the radius is the same.
Add other ifs to assure behaviour equality, or instead of overloading == use getId() to compare it

JPetOptions/JPetOptions.cpp
```diff
@@ -64,8 +64,11 @@ JPetOptions::FileType JPetOptions::handleFileType(const std::string &fileType) c
     }
     catch(const std::out_of_range &outOfRangeFileTypeException)
     {
-      std::string errorMessage = "Out of range error in fileType container ";
-      handleErrorMessage(errorMessage, outOfRangeFileTypeException);
+      // ignore unknown file extension
+      // as the user should be able to define their own extensions
+      //
+      /* std::string errorMessage = "Out of range error in fileType container "; */
```
wkrzemien: remove commented code?

JPetPM/JPetPM.h
```diff
@@ -65,8 +65,8 @@ class JPetPM: public TNamed
   void setBarrelSlot(JPetBarrelSlot &p_barrelSlot){ fTRefBarrelSlot = &p_barrelSlot; }
   JPetBarrelSlot& getBarrelSlot() const { return (JPetBarrelSlot&)*(fTRefBarrelSlot.GetObject()); }	
   
-  inline bool operator==(const JPetPM& pm) { return getID() == pm.getID(); }
-  inline bool operator!=(const JPetPM& pm) { return getID() != pm.getID(); }
+  inline bool operator==(const JPetPM& pm) const { return getID() == pm.getID(); }
```
wkrzemien: The same comment about == overloading as above

JPetPM/JPetPM.h
```diff
@@ -65,8 +65,8 @@ class JPetPM: public TNamed
...
+  inline bool operator!=(const JPetPM& pm) const { return getID() != pm.getID(); }
   
   /*std::vector<TRef> getTRefKBs() const { return fTRefKBs; }
```
wkrzemien: remove commented code

JPetUnpacker/Unpacker2/Unpacker2.cc
```diff
@@ -295,7 +295,9 @@ void Unpacker2::DistributeEvents(string f) {
     Event* event = 0;
     
     // open a new file
-    string newFileName = f + ".root"; 
+    string newFileName = f;
+    newFileName = newFileName.substr(0, newFileName.size() - 3);
```
wkrzemien: string newFileName = f.substr(0, f.size() - 3);

JPetHLDReader/JPetHLDReader.h
```diff
@@ -75,6 +75,11 @@ class JPetHLDReader : public JPetReaderInterface
     else return false;
   }
 
+  virtual TObject* getObject(const char* name){
```
wkrzemien: better getObjectFromFile() ? Im not sure

JPetRawSignal/JPetRawSignal.cpp
```diff
@@ -64,15 +64,15 @@ std::vector<JPetSigCh> JPetRawSignal::getPoints(
 }
 
 
-const JPetSigCh& JPetRawSignal::getTOTPoint() const {
-  return fTOTPoint;
-}
+/* const JPetSigCh& JPetRawSignal::getTOTPoint() const { */
```
wkrzemien: remove commented code

JPetRawSignal/JPetRawSignal.cpp
```diff
@@ -64,15 +64,15 @@ std::vector<JPetSigCh> JPetRawSignal::getPoints(
...
+/* } */
 
-void JPetRawSignal::setTOTPoint(const JPetSigCh & totSigCh) {
-  if (totSigCh.getType() == JPetSigCh::Charge) {
-    fTOTPoint = totSigCh;
-  }
-}
+/* void JPetRawSignal::setTOTPoint(const JPetSigCh & totSigCh) { */
+/*   if (totSigCh.getType() == JPetSigCh::Charge) { */
+/*     fTOTPoint = totSigCh; */
```
wkrzemien: the same

JPetRawSignal/JPetRawSignal.cpp
```diff
@@ -95,6 +95,35 @@ std::map<int, double> JPetRawSignal::getTimesVsThresholdValue(JPetSigCh::EdgeTyp
...
+	map[ it1->getThreshold() ] = it2->getValue() - it1->getValue();	
+      }
+    }
+  }
+
+  return map;
 }
+
+
+/* double JPetRawSignal::getTOT() const { */
```
wkrzemien: remove commented code

JPetAuxilliaryData/JPetAuxilliaryData.cpp
```diff
@@ -0,0 +1,85 @@
...
+JPetAuxilliaryData::JPetAuxilliaryData(std::string name):
+  TNamed(name.c_str(), name.c_str())
+{
+}
+
+void JPetAuxilliaryData::createVector(std::string name, int initial_size){
+  fVectors[name].resize(initial_size, Unset);
+}
+
+void JPetAuxilliaryData::createMap(std::string name){
```
wkrzemien: What if the map already exists?

JPetOptions/JPetOptions.cpp
alekgajos: This will have to be fixed in the longer run - the list of possible file types (eg. phys.hit) is kept in an enum and cannot be easily extended by the user. The above code was commented out to avoid errors related to usage of arbitrary file types created by the user (eg. phys.hit.filtered). Once we allow for such arbitrary file extensions, the out_of_range exception should not happen in sane cases and this code can be uncommented back.

JPetAuxilliaryData/JPetAuxilliaryData.cpp
alekgajos: I understand the problem but I don't know what would be the best solution. I could remove the createMap method completely and it would simply be created at the first call to setvalue("map name",...).

JPetHLDReader/JPetHLDReader.h
alekgajos: done

JPetLayer/JPetLayer.h
alekgajos: I am against removing the comparison operators as the user should not be required to know which field to use to compare parametric objects.

I agree with your point in terms of programming, but we have additional assumptions on the form of such objects from the DB which gives me some concerns:
* in general we should never have two parametric objects with same ID and different properties
* if that happens anyway, it would indicate an error in the DB or in the ParamGetter; then we should signal an error rather than simply return false from the comparison because such comparison makes no logical sense for ill-defined objects.

For now, I have added assertions which check other properties of the IDs are identical.

JPetOptions/JPetOptions.cpp
wkrzemien: Yeah, but since we have the repository we can always come back to it later. The common experience is that the commented code survives suprisingly long, till nobody knows what to do with it. You should remove it now.

JPetAuxilliaryData/JPetAuxilliaryData.cpp
wkrzemien: Well, maybe at least check if the key already exists, if it is the case log it as warning ?
And maybe add a commnet in the .h - that if the map already exists it will be overwritten

JPetTask/JPetTask.h
```diff
@@ -32,16 +33,18 @@ class JPetTask: public JPetTaskInterface, public TNamed
...
   JPetStatistics & getStatistics();
+  JPetAuxilliaryData & getAuxilliaryData();
   virtual TNamed* getEvent() {return fEvent;}
 
  protected:
   TNamed* fEvent;
   JPetParamManager* fParamManager;
   JPetStatistics * fStatistics;
-  
+  JPetAuxilliaryData * fAuxilliaryData;
```
wkrzemien: You forgot to add this member to the initialization list in constructors.

JPetTaskIO/JPetTaskIO.h
```diff
@@ -65,6 +67,7 @@ class JPetTaskIO: public JPetTaskInterface
   JPetReaderInterface* fReader;
   JPetTreeHeader* fHeader;
   JPetStatistics* fStatistics;
+  JPetAuxilliaryData * fAuxilliaryData;
```
wkrzemien: I think there is something wrong in our class hierarchy, 
TaskIO is not a kind of Task? there is a obvious repetition in code, cause part of the members are doubled in both classes

wkrzemien: I think that you should extract JPetStatistics, JPetAuxilliaryData and maybe JPetTreeHeader to a separate class that should be then added to JPetTask and JPetTaskIO as a member. You will avoid the repetition, and make the common interface to it
All three are functionally similar, if I understand correctly. Also, the JPetAuxillaryData seems to repeat part of the functionality that is already in JPetTreeHeader (e.g. dictionnary), maybe one can reuse part of the code

alekgajos: Concerning the class hierarchy - yes, JPetTaskIO is another mplementation of the JPetTaskInterface and not descendant from JPetTask. Anyway, I don't think we should discuss it as a part of this pull request.

wkrzemien: Ok. but in JPetTask fStatisticsis is still not initialized in the default constructor. So fix this at least.

PR #16, opened 2016-07-18
Activate option to unpack first n events from the hld file

wkrzemien:
Also, small code corrections (adding inline)

### DIFF
--- a/JPetTaskExecutor/JPetTaskExecutor.cpp
+++ b/JPetTaskExecutor/JPetTaskExecutor.cpp
@@ -20,6 +20,7 @@
 #include "../JPetTaskLoader/JPetTaskLoader.h"
 #include "../JPetParamGetterAscii/JPetParamGetterAscii.h"
 #include "../JPetParamGetterAscii/JPetParamSaverAscii.h"
+#include "../JPetLoggerInclude.h"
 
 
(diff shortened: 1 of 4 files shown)


JPetTaskExecutor/JPetTaskExecutor.cpp
```diff
@@ -90,7 +91,13 @@ void JPetTaskExecutor::processFromCmdLineArgs(int)
   if (inputFileType == JPetOptions::kScope) {
     createScopeTaskAndAddToTaskList();
   } else if (inputFileType == JPetOptions::kHld) {
-    fUnpacker.setParams(fOptions.getInputFile());
+    long long nevents = fOptions.getTotalEvents();
+    if (nevents > 0) {
+      fUnpacker.setParams(fOptions.getInputFile(), nevents);
+      WARNING(std::string("Even that range of events was set, only the first ")+ JPetCommonTools::intToString(nevents) + std::string(" will be unpacked by the unpacker. \n the unpacker always starts from the beginning of the file"));
```
alekgajos: In the message I would change "Even that range" to "Even though the range".

wkrzemien: ok. changed

PR #17, opened 2016-07-18
Add check if the run number is given from the command line options

wkrzemien:
If not, the false is returned. The only difference is if the scope
type of input file is specified. Then, the run number is ignored by
default. Also, exceptions are thrown in case of not correct options.
A bunch of tests was added.


JPetCmdParser/JPetCmdParser.cpp
```diff
@@ -131,6 +131,13 @@ bool JPetCmdParser::areCorrectOptions(const po::variables_map& variablesMap) con
       return false;
     }
   }
+
+  /// The run number option is neclegted if the input file is set as a  scope file
+  if (isRunNumberSet(variablesMap)) {
+    if (getFileType(variablesMap) =="scope") {
+      WARNING("Run number was specified but the input file type is a scope!\n The run number will be ignored!");
```
alekgajos: "is a scope" -> "is \"scope\"" will be less confusing.

wkrzemien: done

PR #19, opened 2016-07-26
Refactor adding factories

timorl:
Now the creaton of parameter resources is handled by factories
specific for every type of resource. ParamGetters are heavily
simplified and only return plain data.

This change is mostly backwards compatible. Because of that
ParamManager behaves inconsistently -- the new interface
returns different references than the old one.

There are some problems with missing database fields.
Workarounds are in place, hopefully temporarily.

This does not work with current examples -- root fails to generate
some dictionaries, because of the c++11 'using' directives in
one of the headers. Since we might be moving to
ROOT 6 soon [citation needed]  I didn't want to fix that yet.


timorl:
Rebased after latest changes. Had to use a try-catch block in one place, removing it would be tricky.

Also the last paragraph in the original message is no longer valid -- the offending "using" directives have been changed to "typedef".

PR #20, opened 2016-09-05
Fix progressbar. Change related unit tests.

andprogrammer:
Value after -b option is no longer required.
Pull request j-pet-framework : andprogrammer (master) -> JPETTomography (master).

### DIFF
--- a/JPetTaskIO/JPetTaskIO.cpp
+++ b/JPetTaskIO/JPetTaskIO.cpp
@@ -174,9 +174,9 @@ void JPetTaskIO::createInputObjects(const char* inputFilename)
   }
 }
 
-void JPetTaskIO::createOutputObjects(const char* outputFilename)
+void JPetTaskIO::createOutputObjects(const char* outputFilename, const char* outputFileDirectory)
 {
-  fWriter = new JPetWriter( outputFilename );
+  fWriter = new JPetWriter( outputFilename, outputFileDirectory );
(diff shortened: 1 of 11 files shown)


wkrzemien:
You should change your PR target to j-pet-framework/develop. In our workflow we never merge directly to master. 
The second thing -> I don't like the idea to override the methods everywhere by adding second argument - since it breaks many interfaces e.g. 
I would rather add the path to the filename, and pass it as a whole - > and it should work, you dont need to change the methods everywhere.

JPetTaskIO/JPetTaskIO.cpp
```diff
@@ -174,9 +174,9 @@ void JPetTaskIO::createInputObjects(const char* inputFilename)
   }
 }
 
-void JPetTaskIO::createOutputObjects(const char* outputFilename)
+void JPetTaskIO::createOutputObjects(const char* outputFilename, const char* outputFileDirectory)
 {
-  fWriter = new JPetWriter( outputFilename );
+  fWriter = new JPetWriter( outputFilename, outputFileDirectory );
```
wkrzemien: Instead of adding the second argument, you can merge both arguments in one string
and use the old version of the writer.

andprogrammer:
Reopen one another pull request has been opened. Based on develop branch. New PR contains changes due to progressbar only.

PR #30, opened 2016-10-11
Migration to ROOT6

Alvarness:
ROOT 6 compiled by function:
cmake -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0" /path_to_root_source/
make

framework prerequisites:
sudo aptitude install libboost-all-dev libpqxx-dev libpqxx-4.0 libconfig++9v5 libconfig++-dev libtinyxml2-dev doxygen
Compiles on clean Linux Mint 18 install without problems.

HOWEVER, tests do not compile - probably due to some problems with libraries linking in cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,30 +16,15 @@ if(NOT MSVC)
   add_definitions(-std=c++11 -Wall -Wunused-parameter)
 endif()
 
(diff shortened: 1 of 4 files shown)


wkrzemien:
I am closing it, since we will not do it in the near future. Might be reaopen in the future

PR #35, opened 2016-10-24
Ready JPetTreeHeader and JPetTimeWindow

kmuzalewska:
JPetTimeWindow do not has duplicate methods. It was also corrected in the tests.
JPetTreeHeader has corrected methods (getVariable() and getProcessingStageInfo). This class has added JPetTreeHeaderTests.


JPetTreeHeader/JPetTreeHeader.cpp
```diff
@@ -120,5 +122,6 @@ void JPetTreeHeader::setVariable(std::string name, std::string value){
  *
  */
 std::string JPetTreeHeader::getVariable(std::string name) const {
-  return fDictionary.at(name); 
+	if(fDictionary.find(name)!= fDictionary.end()) return fDictionary.at(name);
+	else return " ";
```
wkrzemien: You should return ""  and not " ". Technically " " is not an empty string. Empty string has a length of 0.
Then change the tests to fullfili it.

PR #39, opened 2016-11-16
Add zip option

wictus:
First attempt to add zip option. Gz compressed files will come at this point only in hld format.
Gz file is first unpacked by from the system, then unpacker is launched.


JPetTaskExecutor/JPetTaskExecutor.cpp
```diff
@@ -118,6 +120,28 @@ bool JPetTaskExecutor::processFromCmdLineArgs(int)
     }
     unpackFile();
   }
+  else if( inputFileType == JPetOptions::kZip){
+    INFO( std::string("Unzipping file before unpacking") );
+    unzipFile();
+    //Changing file name to proper one after unzipping
```
wkrzemien: Extract a function that does  or probably have a look if we dont have already some function that changes the suffixes.

JPetTaskExecutor/JPetTaskExecutor.cpp
```diff
@@ -142,6 +166,17 @@ void JPetTaskExecutor::unpackFile()
   }
 }
 
+void JPetTaskExecutor::unzipFile()
```
wictus: System returns what "gzip" will return. 0 in case of success and 1 if something went wrong.

JPetCommonTools/JPetCommonToolsTest.cpp
```diff
@@ -140,4 +140,15 @@ BOOST_AUTO_TEST_CASE(appendSlashToPathIfAbsent)
   BOOST_REQUIRE_EQUAL(JPetCommonTools::appendSlashToPathIfAbsent("test"), "test/");
 }
 
+BOOST_AUTO_TEST_CASE(tryToUnzipSomethingNotExistingFile)
+{
+  BOOST_REQUIRE_EQUAL(JPetCommonTools::unzipFile("kiko.gz"), true);
```
wkrzemien: if it is true or false
you can use 
BOOST_REQUIRE(...)

JPetCommonTools/JPetCommonToolsTest.cpp
```diff
@@ -140,4 +140,15 @@ BOOST_AUTO_TEST_CASE(appendSlashToPathIfAbsent)
...
+  std::string initialPath= boost::filesystem::path(boost::filesystem::current_path()).string();
+  initialPath = initialPath.substr(0, initialPath.find("build") );
+  std::string wrongZipPath = initialPath + "j-pet-framework/unitTestData/JPetCommonToolsTest/wrongZip.gz";
+  BOOST_REQUIRE_EQUAL(JPetCommonTools::unzipFile( wrongZipPath.c_str() ), true);
```
wkrzemien: The same here

JPetCommonTools/JPetCommonToolsTest.cpp
```diff
@@ -140,4 +140,15 @@ BOOST_AUTO_TEST_CASE(appendSlashToPathIfAbsent)
   BOOST_REQUIRE_EQUAL(JPetCommonTools::appendSlashToPathIfAbsent("test"), "test/");
 }
 
+BOOST_AUTO_TEST_CASE(tryToUnzipSomethingNotExistingFile)
+{
+  BOOST_REQUIRE(JPetCommonTools::unzipFile("kiko.gz"));
+  std::string initialPath= boost::filesystem::path(boost::filesystem::current_path()).string();
+  initialPath = initialPath.substr(0, initialPath.find("build") );
+  std::string wrongZipPath = initialPath + "j-pet-framework/unitTestData/JPetCommonToolsTest/wrongZip.gz";
+  BOOST_REQUIRE(JPetCommonTools::unzipFile( wrongZipPath.c_str() ));
```
wkrzemien: Why it returns true? shouldn't it return false when the path is wrong?

JPetCommonTools/JPetCommonTools.h
```diff
@@ -123,6 +123,10 @@ class JPetCommonTools : public boost::noncopyable
   inline static bool isDirectory( const std::string& dir) {
     return boost::filesystem::is_directory(dir);
   }
+  
+  inline static bool unzipFile(const char* filename){
+    return system( Form("gzip -d %s", filename ) );
```
wkrzemien: Could you change to: (std::string("gzip -d ") +filename).c_str()
so we dont need to use Form

JPetCommonTools/JPetCommonToolsTest.cpp
wictus: It returns 0 if everything is correct and error code when not. So from point of booleans it is false when everything went ok, and true if not ...

JPetCommonTools/JPetCommonTools.h
```diff
@@ -30,8 +30,8 @@ class boost::noncopyable;
 #include <iomanip>
 #include <algorithm>
 #include <cctype>
-
-
+#include <stdlib.h>
+#include <TString.h>
```
wkrzemien: And this can be removed if you change the thing below

JPetTaskExecutor/JPetTaskExecutor.cpp
```diff
@@ -104,20 +104,27 @@ bool JPetTaskExecutor::processFromCmdLineArgs(int)
...
-    } else {
-      fUnpacker.setParams(fOptions.getInputFile());
-    }
-    unpackFile();
+    unpackFile(inputFile, nevents);
   }
+  //Assumption that only HLD files can be zipped!!!!!! - Sz.N.
+  else if( inputFileType == JPetOptions::kZip){
+    INFO( std::string("Unzipping file before unpacking") );
+    JPetCommonTools::unzipFile(inputFile);
```
wkrzemien: you should write if (...) and log ERROR(...)  if it is not ok.

JPetCommonTools/JPetCommonToolsTest.cpp
wictus: Ok. I will change it.

JPetCommonTools/JPetCommonTools.h
```diff
@@ -123,6 +123,10 @@ class JPetCommonTools : public boost::noncopyable
   inline static bool isDirectory( const std::string& dir) {
     return boost::filesystem::is_directory(dir);
   }
+  
+  inline static bool unzipFile(const char* filename){
```
wkrzemien: maybe add negation because it will be less suprising. And change the tests accrodingly

PR #40, opened 2016-11-18
Changed method names from getId() to getID()

wictus:


wkrzemien:
Something is wrong now you have 15 files changed and it should be only 3, max 4 (JPetLayer , JPetLayerTest )?
+ there is some conflict

PR #42, opened 2016-11-28
JPetEvent data class.

wkrzemien:
PetEvent data class  should be more general than JPetLOR, because it accepts different number of hits. Added some tests, and a method that returns hits ordered by time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,7 @@ generate_root_dictionaries(DICTIONARIES SOURCES ${SOURCES}
   JPetFEB
   JPetUser
   JPetUnpacker
+  JPetEvent
   )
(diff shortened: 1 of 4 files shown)


alekgajos:
I am not sure about two things (but I myself don't know which solution is better):
- whether of not the event should have a single time (and its quality); it makes sense once this is a reconstructed annihilation event but we may as well want to store hits from more than one annihilation (e.g. before we determine what kind of event it was it may not be known how many physical events with their times are inside); perhaps we should just treat this Event as a container for hits?
- it might be good to add some way for the user to store information about the type of event (2g, 3g, 2g+prompt, 3g+prompt, 2g with scattering etc.); perhaps some enum made with bitmasks so that types can be added like type = 2G | PROMPT |  SCATTER ?

Alvarness:
Also, hits needs to be ordered by time before dividing them into events.

PR #44, opened 2016-11-28
Unifying names of method getID() in JPetLayer and JPetPMCalib 

wictus:


wkrzemien:
I'm closing cause the new version #45  has no conflicts

PR #47, opened 2016-12-03
JPetEvent data class should be more general than JPetLOR, because it accepts different number of hits. Added some tests, and a method that returns hits ordered by time. (reopen)

wkrzemien:
I am opening it back, because JPetEvent classes were removed by accident by accepting PR #45 .
I also copy two comments below. We should agree on it before merging:
@alekgajos 
"I am not sure about two things (but I myself don't know which solution is better):

whether of not the event should have a single time (and its quality); it makes sense once this is a reconstructed annihilation event but we may as well want to store hits from more than one annihilation (e.g. before we determine what kind of event it was it may not be known how many physical events with their times are inside); perhaps we should just treat this Event as a container for hits?
it might be good to add some way for the user to store information about the type of event (2g, 3g, 2g+prompt, 3g+prompt, 2g with scattering etc.); perhaps some enum made with bitmasks so that types can be added like type = 2G | PROMPT | SCATTER ?"

@Alvarness :
"Also, hits needs to be ordered by time before dividing them into events."


wkrzemien:
@Alvarness @alekgajos 
I think the method that returns hits  ordered in time its enough.  Or if we think that really in all cases we need hits ordered by time I will add ordering in the constructor and set method. I don't want to rely on users providing correct order - its  error-prone.

Definitely the FLAG with event would be useful. Could you propose any classification? Even for now?
Concerning the time of anihiliation I have no opinion. I can remove it

Alvarness:
In order to divide hits into events they are most likely should be ordered by time - however in fact it might be enough to order them i task at the level of saving hits rather than as a method of class.

Also I think that we need two different event classes - one for preselected hits (that we think might come from one event - but might be from several) and one for actual events (so after some analysis performed on previous event class) with flag.

As for the event classification I think that four types are sufficient - 2g/3g with or without deexcitation photon.

wkrzemien:
@Alvarness 
If the events (before and after preselection) had exactly  same signature (methods and so on) , then it would be enough to add to the enum list a type preselected without adding two different class types. Then in one module you could create kPRESELECTED events, and in the next one create already actual events with different enum (2g, 3g etc). 

I don't know if I got your first comment correctly. I see two options: or assure hit order while constructing or setting hits (so effectively in the JPetEvent constructor ans setHits method) or leave the function that return ordered hits.  What I dont want is to rely on user that he maybe order hit correctly, even that would be necessary in previous modules.

alekgajos:
Concerning hits order: I would go for assuring that the hits provided to the event are already sorted. I agree that practically it would hardly make sense to group hits into some events without having them ordered first.

As for the two stages (preselection and real physical events) - it is true that we probably will need both these steps. Maybe two classes are not a bad idea. Then this JPetEvent could be a container for (time-ordered) hits without any particular time information but with an additive classification flag indicating anything classifiable that was seen in this time window like:
* 2gamma
* 3gamma
* prompt
* scattering
* (maybe it would be reasonable to have some flags for unexpected use like USER1, USER2 etc. ?)

The second class would represent a 2g/3g event with optional information about associated prompt gamma and the following information: time, tof, prompt-time, origin-point, etc.

Otherwise we can do with one class only and introduce all this information like tof, optional prompt time etc. into the JPetEvent but prevent it from being accessed unless the event is clearly classified as e.g. 2gamma|prompt.

wkrzemien:
OK,  we can always overload the JPetEvent, or add a pointer to some struct with those additional info.
But it would be good to define already what we need if we want to add the second one
I added newer version of JPetEvent.

wkrzemien:
I will add enums later

Alvarness:
After some considerations - we need two containers with different methods.

First will store hits divided based on some cuts (most likely time). We need to determine number of events from which that hits comes (by a built-in method?). Also return (save) that events. This is basically everything that class needs.

Second one store hits from single events. As a method it should return type of event (2g/3g with or without prompt gamma - scattering is irrelevant - it can be determined based on number of hits). Also point of annihilation (maybe time of annihilation) and some other methods.

wkrzemien:
Added JPetEventType enum type and small corrections for JPetHit class.

PR #50, opened 2016-12-10
protected structures fix: they were made public + Time synchronization classes were added

alexkernphysiker:
there was compile error in dictionary files


wkrzemien:
You also added LargeBarrelExtensions  with this PR - do you want to add now? - then change the title pls -?

LargeBarrelExtensions/BarrelExtensions.cpp
```diff
@@ -0,0 +1,141 @@
+#include <algorithm>
+#include <sstream>
+#include "BarrelExtensions.h"
+using namespace std;
+const string Layer(const size_t layer){
+    stringstream stream;
```
wkrzemien: In  c++11 there is to_string method in the standard library, so you should use it

(By the way we also should do it in other part of the framework, we had some tools in JPetCommonTools, now it becomes obsolete)

LargeBarrelExtensions/BarrelExtensions.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return stream.str();
+}
+const string LayerSlot(const size_t layer, const size_t slot){
+    stringstream stream;
+    stream<<"layer"<<layer<<"-slot"<<slot;
+    return stream.str();
+}
+const string LayerSlotThr(const size_t layer, const size_t slot, const size_t thr){
+    stringstream stream;
+    stream<<"layer"<<layer<<"-slot"<<slot<<"-thr"<<thr;
```
wkrzemien: the same as above

LargeBarrelExtensions/BarrelExtensions.h
```diff
@@ -0,0 +1,76 @@
+#ifndef _LARGE_BARREL_EXTENSIONS_
```
wkrzemien: licence header is missing

alexkernphysiker:
I first made a pull request, than I committed new changes to my repository. what am I doing wrong?

LargeBarrelExtensions/BarrelExtensions.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return {.layer=getLayerNumber(slot.getLayer()),.slot=getSlotNumber(slot)};
+}
+const vector< size_t > AbstractBarrelMapping::getLayersSizes() const{
+    vector<size_t> res;
+    for(size_t i=1,n=getLayersCount();i<=n;i++)
+	res.push_back(getSlotsCount(i));
+    return res;
+}
+
+const double degFullCircle = 360.;
```
wkrzemien: Do you need this global?
btw we have a convention to start consts with kDegFullCircle

alexkernphysiker: It was just taken from Alek's example. OK, I'll change it

LargeBarrelExtensions/BarrelExtensions.cpp
wkrzemien: It is shorter and more expressive:
return "layer" + to_string(layer);

wkrzemien:
If  you add new changes to the same branch on your github account it automatically updates PR. you must create a separate branch if you want to add some new stuff and not affect the PR

LargeBarrelExtensions/BarrelExtensions.cpp
alexkernphysiker: OK. I'll yield to the dictation. Nevertheless, I don't think that this one is less expressive

LargeBarrelExtensions/BarrelExtensions.cpp
alexkernphysiker: :laughing:  It even wasn't used. Deleting!

wkrzemien:
-You should change methods names to small letters (now it is a mixture of both cases - sometimes small, sometimes capital).
-Add some description to Synchronization class.
-And there is some conflict in CMakeLists.txt

alexkernphysiker:
Which conflict do you mean?

PR #52, opened 2016-12-13
Internal refactoring

wkrzemien:
Please have a look @alekgajos 
This is reopened PR #41 
Those changes should be transparent to Framework users.
The main changes:
1.Rename JPetTaskExecutor to JPetTaskChainExecutor.
2. The role of the JPetTaskChainExecutor is to execute task chain, all other "things" are delegated to JPetTaskChainExecutorUtils via preprocess() method. JPetTaskChainExecutorUtils contains all the "preprocessing tasks" e.g. unzipping, unpacking, generateParamManager (from file)
3.Add JPetOptionInterface -> JPetOptions is now a subtype of it
4.Small corrections e.g. missing initializations in JPetTimeWindow, JPetHLDReader, remove unused code, formatting
5.Remove inheritance JPetTaskInterface <--JPetTaskIO, and partially change the inheritance scheme that was bad anyway. We have two type of classes with Task in name which has nothing to do with each other. JPetTask is a "real" task (computing component/algorithm), while JPetTaskIO, JPetTaskLoader are more like manager which handle the tasks.
6.Refactor our task scheme to introduce properly Strategy pattern. It was implemented but not correctly (probably by accident).
See e.g. https://en.wikipedia.org/wiki/Strategy_pattern
JPetTaskRunner is a context of the strategy pattern. JPetTaskInterface corresponds to the Strategy interface and JPetTasks are concrete realizations of different Strategies.

This is the preparation for the second part of refactoring, but this will break some user interfaces, so I leave it for the future.

### DIFF
--- a/JPetCommonTools/JPetCommonTools.h
+++ b/JPetCommonTools/JPetCommonTools.h
@@ -48,9 +48,9 @@ public:
     out << x;
     return out.str();
   }
-  
+
   static std::string doubleToString(double x);
-  
+
   static int stringToInt(const std::string& str);
 
   static bool to_bool(std::string str) {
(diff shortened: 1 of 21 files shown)


alekgajos:
I have no further comments and I can confirm that examples work fine with this version.

PR #56, opened 2016-12-15
Easy choosing of different thresholds for time synchronization

alexkernphysiker:


JPetLargeBarrelExtensions/TimeSyncDeltas.cpp
```diff
@@ -36,6 +36,13 @@ const SynchroStrip Synchronization::get_times(const JPetHit & hit) const{
     res.B=f_time_calc(B)+offsets.B;
     return res;
 }
-const double defaultTimeCalculation(const vector<double>&P){
-    return P[0];//This algorithm should be discussed
+
+const Synchronization::TimeCalculation Thr(const std::list<size_t>&lst){
+    return [lst](const vector<double>&P)->const double{
```
wkrzemien: I would check if list.size is not 0 since you are dividing by it later
and maybe call lst    thresholds is more meaningful
But those are minor things. As you prefer.

PR #57, opened 2016-12-29
Create sinogram from emission matrix

grey277:
Tests are not completed, for phantom test, phantom.pgm file is required in tests folder, should I add it by git, hardcode it in test file(it is just string with value of every pixel in greyscale) or add it by any other way? Also I added modules/JPet*/*Test.cpp to CMake UNIT_TEST_SOURCES variable for making tests from modules(maybe it should be done in other way?).


wkrzemien:
Ok. Great. I will have a look at it. The test file, should be copied to our server. Send me the file by mail and I will do it. It will be automatically put to j-pet-framework/unitTestData/JPetSinogram  when calling cmake.
See e.g. j-pet-framework/unitTestData/JPetReader/ and JPetReaderTest.cpp 
It sets the path to "unitTestData/JPetReaderTest/timewindows.root", so you should do analogically.

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+  } i=0;
+
+  int scanNumber=0;
+  int inputMatrixSize = emissionMatrix.size1();
+  int x = 0, y = 0, Xcenter = 0, Ycenter = 0;
+  Xcenter = inputMatrixSize / 2;
+  Ycenter = inputMatrixSize / 2;
+
+  //if no. scans is greater than the image width, then scale will be <1
+  double scale = inputMatrixSize*1.42/scans;
```
wkrzemien: what is 1.42 ?
Anyway dont use magic number. At least change it to some const and give some meaningful name or/and command

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+  
+  int N=0;
+  double value = 0.;
+  double weight = 0.;
+  for (phi=ang1;phi<ang2;phi=phi+stepsize){
+      double a = -costab[i]/sintab[i];
+      double aa = 1/a;
```
wkrzemien: is it guaranteed that a !=0 ?

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+JPetSinogram::JPetSinogram() { }
+
+JPetSinogram::~JPetSinogram() { }
+
+/*! \brief Function returning vector of vectors with value of sinogram(180/views degree step and emissionMatrix.size1()/scans size) scaled to <min,max> 
+ *  \param emissionMatrix matrix, need to by NxN
+ *  \param views number of views on object, degree step is calculated as (ang2 - ang1) / views
+ *  \param scans number of scans on object, step is calculated as emissionMatrix.size(1) / scans
+ *  \param fast if true use nearest neighbour interpolation instead linear interpolation in calculation (Optional, default false)
```
wkrzemien: I would suggest changing parameter "fast" to 
"nearest".

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+  assert(min < max);
+  assert(ang1 < ang2);
+
+  //create vector of size views, initialize it with vector of size scans
+  std::vector<std::vector<double>> proj(views, std::vector<double>(scans)); 
+  std::unique_ptr<double[]> sintab(new double[views]);
+  std::unique_ptr<double[]> costab(new double[views]);
+  
+  float phi = 0., stepsize = 0.;
+  stepsize = (ang2 - ang1) / views;
```
wkrzemien: float stepsize =...

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+  assert(scans > 0);
+  assert(min < max);
+  assert(ang1 < ang2);
+
+  //create vector of size views, initialize it with vector of size scans
+  std::vector<std::vector<double>> proj(views, std::vector<double>(scans)); 
+  std::unique_ptr<double[]> sintab(new double[views]);
+  std::unique_ptr<double[]> costab(new double[views]);
+  
+  float phi = 0., stepsize = 0.;
```
wkrzemien: no need to define stepsize here

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+          for (scanNumber=0;scanNumber<scans;scanNumber++){
+              N = scanNumber - scans/2;
+              double b = (N - costab[i] - sintab[i]) / sintab[i];
+              b =  b * scale;
+              
+              for (x = -Xcenter; x < Xcenter; x++){
+                  if (fast == true){ //nearest neighbour interpolation
+                      y = (int) std::round(a*x + b);
+                      
+                      if (y >= -Xcenter && y < Xcenter )
```
wkrzemien: Just for clarity I would change this argument to : if (y >= -Ycenter && y < Ycenter) 
I understand that it would be the same, but it is easier to understand

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+                          value += emissionMatrix((x+Xcenter),(y+Ycenter));
+                      
+                  } else {
+                      y = (int) std::round(a*x + b); //linear interpolation
+                      weight = std::abs((a*x + b) - std::ceil(a*x + b));
+                      
+                      if (y >= -Xcenter && y+1 < Xcenter )
```
wkrzemien: As before would be better to compare with  Ycenter

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+  for (phi = ang1; phi < ang2; phi = phi + stepsize) {
+      sintab[i] = std::sin((double) phi * M_PI / 180 - M_PI/2);
+      costab[i] = std::cos((double) phi * M_PI / 180 - M_PI/2);
+      i++;
+  } i=0;
+
+  int scanNumber=0;
+  int inputMatrixSize = emissionMatrix.size1();
+  int x = 0, y = 0, Xcenter = 0, Ycenter = 0;
+  Xcenter = inputMatrixSize / 2;
```
wkrzemien: const int Xcenter = inputMatrixSize/2;
const int Ycenter = inputMatrixSize/2;

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+                  }
+              } proj[i][scanNumber] = value/std::abs(costab[i]); value=0;
+              
+          }
+          
+      } i++;
+  }
+  i=0;
+  if(scaleResult) {
+    double datamax = proj[0][0];
```
wkrzemien: extract the scale procedure into a separate function

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
+#include "./JPetSinogram.h"
+#include <boost/numeric/ublas/matrix.hpp>
+#include <vector>
+#include <utility>
+#include <cmath>
+using namespace boost::numeric::ublas;
+
+#define sang std::sqrt(2)/2
```
wkrzemien: you dont need this to be global define it only inside the function, as const double

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+JPetSinogram::JPetSinogram() { }
```
wkrzemien: I know I think I suggested it. But I would change the name of this class to JPetRecoImageTools or something like that.

Also, add private copy constructor and =operator to it.
or:
JPetRecoImageTools(const JPetRecoImageTools&) = delete;
JPetRecoImageTools& operator=(const JPetRecoImageTools&) = delete;

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+                      if (y >= -Xcenter && y+1 < Xcenter )
+                          value += (1-weight) * emissionMatrix((x+Xcenter),(y+Ycenter))
+                                  + weight * emissionMatrix((x+Xcenter), (y+Ycenter+1));
+                      
+                  }
+              } proj[i][scanNumber] = value/std::abs(sintab[i]); value=0;
+              
+          }
+      }
+      if (std::abs(sintab[i]) <= sang){
```
wkrzemien: this part of the code is almost a repetition of the upper part, so you should refactor it e.g. you can extract all the code that is the same leaving if (std::abs() <= sang) condition with the things that really change, or you can even define some helper function and put all the common operation there. It up to you.

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+      double aa = 1/a;
+      if (std::abs(sintab[i]) > sang){
+          for (scanNumber=0;scanNumber<scans;scanNumber++){
+              N = scanNumber - scans/2;
+              double b = (N - costab[i] - sintab[i]) / sintab[i];
+              b =  b * scale;
+              
+              for (x = -Xcenter; x < Xcenter; x++){
+                  if (fast == true){ //nearest neighbour interpolation
+                      y = (int) std::round(a*x + b);
```
wkrzemien: even that those are only few lines
I would extract lines 74-78 and 80 to 86 as two interpolation methods,
You can even pass that function as std::function argument to your sinogram method and remove the condition  if (fast==true) totally.

modules/JPetSinogram/JPetSinogram.cpp
```diff
@@ -0,0 +1,140 @@
...
+      double a = -costab[i]/sintab[i];
+      double aa = 1/a;
+      if (std::abs(sintab[i]) > sang){
+          for (scanNumber=0;scanNumber<scans;scanNumber++){
+              N = scanNumber - scans/2;
+              double b = (N - costab[i] - sintab[i]) / sintab[i];
+              b =  b * scale;
+              
+              for (x = -Xcenter; x < Xcenter; x++){
+                  if (fast == true){ //nearest neighbour interpolation
```
wkrzemien: if (fast)

modules/JPetSinogram/JPetSinogramTest.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+    std::cout << std::endl;
+  }
+}
+
+BOOST_AUTO_TEST_CASE(sinogram)
+{
+
+  std::ifstream in;
+  in.open("unitTestData/JPetSinogram/phantom.pgm");
```
wkrzemien: Sorry my bad -should be unitTestData/JPetSinogramTest/phantom.pgm"

PR #59, opened 2017-01-03
Raw signal remove duplicated method

Alvarness:


wkrzemien:
Please, remove also the corresponding lines in JPetRawSignalTest.cpp, If not we will get the error.

PR #60, opened 2017-01-03
 Add return dummy class when reference is not set

grey277:
add method .isNullObject() to check is returned reference is dummy

### DIFF
--- a/JPetHit/JPetHit.cpp
+++ b/JPetHit/JPetHit.cpp
@@ -103,20 +103,18 @@ const JPetPhysSignal& JPetHit::getSignalB() const
 }
 const JPetScin& JPetHit::getScintillator() const
 {
-  static JPetScin DummyResult(true);
   if(fScintillator.GetObject()) return (JPetScin&) * fScintillator.GetObject();
   else {
     ERROR("No JPetScin slot set, Null object will be returned");
-    return DummyResult;
+    return JPetScin::getDummyResult();
   }
 }
 const JPetBarrelSlot& JPetHit::getBarrelSlot() const
 {
-  static JPetBarrelSlot DummyResult(true);
   if(fBarrelSlot.GetObject()) return (JPetBarrelSlot&) * fBarrelSlot.GetObject();
   else {
     ERROR("No JPetBarrelSlot slot set, Null object will be returned");
-    return DummyResult;
+    return JPetBarrelSlot::getDummyResult();
   }
 }
 bool JPetHit::isSignalASet() const
--- a/JPetHit/JPetHit.cpp
+++ b/JPetHit/JPetHit.cpp
@@ -23,8 +23,6 @@ ClassImp(JPetHit);
 JPetHit::JPetHit() :
   TNamed("JPetHit", "Hit Structure")
 {
-  fIsSignalAset = false;
-  fIsSignalBset = false;
 }
 
 JPetHit::JPetHit(float e, float qe, float t, float qt, TVector3& pos, JPetPhysSignal& siga, JPetPhysSignal& sigb,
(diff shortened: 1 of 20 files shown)


JPetHit/JPetHit.cpp
```diff
@@ -56,8 +56,16 @@ const JPetPhysSignal& JPetHit::getSignal(Signal pos) const {
 }
 const JPetPhysSignal& JPetHit::getSignalA() const {return fSignalA;}
 const JPetPhysSignal& JPetHit::getSignalB() const {return fSignalB;}
-const JPetScin& JPetHit::getScintillator() const {return (JPetScin&)*fScintillator.GetObject();}
-const JPetBarrelSlot& JPetHit::getBarrelSlot() const {return (JPetBarrelSlot&)*fBarrelSlot.GetObject();}
+const JPetScin& JPetHit::getScintillator() const {
+  static JPetScin DummyResult(true);
+  if(fScintillator.GetObject()) return (JPetScin&)*fScintillator.GetObject();
+  else return DummyResult;
```
alekgajos: I believe that at this point we should already signal an error in the log (and possibly even terminate the program...). 
I would not assume the users will check for these null objects at the level of their code every time they call a getScintillator() method on a hit.

JPetScin/JPetScin.h
```diff
@@ -53,11 +54,17 @@ class JPetScin: public TNamed
   void setScinSize(Dimension dim, float value);
 
   void setBarrelSlot(JPetBarrelSlot &p_barrelSlot){ fTRefBarrelSlot = &p_barrelSlot; }
-  JPetBarrelSlot& getBarrelSlot() const { return (JPetBarrelSlot&)*(fTRefBarrelSlot.GetObject()); }	
+  JPetBarrelSlot& getBarrelSlot() const { 
+    static JPetBarrelSlot DummyResult(true);
+    if(fTRefBarrelSlot.GetObject()) return (JPetBarrelSlot&)*(fTRefBarrelSlot.GetObject());
+    else return DummyResult;
+    }
```
alekgajos: Same comment as for JPetHit.cpp

alekgajos:
As one general remark, since all the parametric objects now have a "isNullObject" member and a respective method, we could consider extracting them to a common base class for all such objects.

wkrzemien:
1.There are some conlficts now, probably due to some changes merged to develop. Please update your version.
2. I agree with Alek's remark,  please add to every else statement
ERROR("No barrel slot set,  Null object will be return"); or something similar.
3.I think you should define  NullObjects as singletons inside the classes they represent. In your representation now you need e.g. a separate static JPetLayer for every getLayer() method in any class. I think that you should have just one static JPetLayer, so you can add to JPetLayer some static method getNullObject() that returns static JPetLayer; or something similar.
4.About base class I dont know it seems nice on the other hand the inheritnace chain is getting complicated, maybe one could use composition

grey277:
Ok. I resolved problem with merge, added ERROR message when returning DummyResult and created method to get DummyClass. I didn't add termination of program when JPetScin is null in JPetHit or JPetBarrelSlot is null in JPetScin.

wkrzemien:
For me it seems fine. Can you add two more things?
1. Increment every ClassDef by one e.g.  ClassDef(JPetFEB, 1); - >ClassDef(JPetFEB, 2); This is internal ROOT mecanism to recognise the version.
2. If possible for all classes move fields initialization from constructor initialization list to class definition.
 fId(-1),   to  fId =-1; 
but check if for all constructors it is ok.

grey277:
Ok, i changed that, all tests are passing without any error, only JPetCmdParserTest can't find unitTestData/JPetCommonToolsTest/goodZip.gz file (i tried deleting build folder and run cmake again to download file, but still I don't have it, maybe it is not added to sphinx server)

wkrzemien:
Ok. Great. I will check the missing file.

PR #61, opened 2017-01-04
Travis

kmuzalewska:
Corrected parserXML. If test fails, travis fails. Additionally parser prints name of test where was a failure.


unitTestUtils/parseXML.py
```diff
@@ -1,21 +1,30 @@
...
+def eprint(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
 
 def parse():
     for infile in glob.glob('*.xml'):
-	try:
+        try:
             tree = ET.parse(infile)
-        except ParseError:
-    	    print "The file xml isn't correct. There were some mistakes in the tests "
```
wkrzemien: since you already imported print as function, please use it properly everywhere
print("The file xml ...")

PR #62, opened 2017-01-05
Refactoring of JPetRecoImageTools

wkrzemien:
Refactoring of JPetRecoImageTools including renaming variables, introducing typedef, and extracting some helper functions. Removing the discontinuity bug
Still some tests are missing (e.g. projections and sinograms).

### DIFF
--- a/modules/JPetRecoImageTools/JPetRecoImageTools.cpp
+++ b/modules/JPetRecoImageTools/JPetRecoImageTools.cpp
@@ -26,7 +26,6 @@ JPetRecoImageTools::JPetRecoImageTools() {}
 
 JPetRecoImageTools::~JPetRecoImageTools() {}
 
-//nearest neighbour
 double JPetRecoImageTools::nearestNeighbour( int i, double y, std::function<double(int, int)>& func)
 {
   int j = std::round(y);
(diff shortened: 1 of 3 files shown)


modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -10,180 +26,173 @@ JPetRecoImageTools::JPetRecoImageTools() {}
...
+  int viewIndex = 0;
+  const double kSin45deg = std::sqrt(2) / 2; /// sin(45) deg
+  for (auto phi = angleBeg; phi < angleEnd; phi = phi + stepsize) {
+    sinValue = std::sin(phi * kDegToRad - M_PI / 2.);
+    sinValue = setToZeroIfSmall(sinValue, kEpsilon);
+    cosValue = std::cos(phi * kDegToRad - M_PI / 2.);
+    cosValue = setToZeroIfSmall(cosValue, kEpsilon);
+
+
     a = -cosValue / sinValue;
```
grey277: I think it should be changed with lines 114 - 120, also then it is no need for extra variable aa(it is just a)

wkrzemien: Right. I will change it.

PR #63, opened 2017-01-07
Test data fetching with checksums

timorl:
Not yet fully tested because of problems with mystical beasts.


timorl:
Should work now. The script now downloads a checksum file, checks all the files in it and downloads the ones that are wrong.

PR #65, opened 2017-01-11
Jpetoptions

kmuzalewska:
problems with JPetCmdParser.cpp. I tried to include options from json before generating options->it is last 'if' in parseandgenerateoptions but can not find solution to problem with type variable_value from boost


old/JPetCmdParserTest/testfile.json
```diff
@@ -0,0 +1,57 @@
+{
```
wkrzemien: All the data test files should be put in the unitTestData/JPetCmdParserTest at the sphinx server. It will be automatically copied by cmake. You can send me the files by mail.

wkrzemien: So you mist remove it from the PR

old/master/JPetUnpackerTest/conf_trb3.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
```
wkrzemien: I think you should remove from this PR all old folder. I dont know what it is for.

JPetCmdParser/JPetCmdParser.cpp
```diff
@@ -61,6 +65,18 @@ std::vector<JPetOptions> JPetCmdParser::parseAndGenerateOptions(int argc, const
   if (!areCorrectOptions(variablesMap)) {
     throw std::invalid_argument("Wrong user options provided! Check the log!");
   }
+  if(variablesMap.count("json")){
```
wkrzemien: For clarity:
1.I would add the method  that returns a map of options  
`static std::map<std::string, std::string> createOptionsFromFile(const std::string& filename)`
because it makes more sense here than createing JPetOptions object
2.I would  extract a method e..g:
 mergeOptions(const& std::map<std::string, std::string> options, out variableMap) {
content from 70 to 77 -> think what to do with the exception
} 
3.I would change 68 to 79 into:
std::map<std::string, std::string> jsonOptions
if(variablesMap.count("json")) {
 auto jsonCfgFile = variablesMap["json"].as<std::string>();
 jsonOptions = createOptionsFromFile(jsonCfgFile);
}
 mergeOptions(jsonOptions, variablesMap);
return generateOptions(variablesMap);

unitTestUtils/parseXML.py
```diff
@@ -1,21 +1,30 @@
 #!/usr/bin/python
```
wkrzemien: the newer version is already included so it is not necessary

JPetCmdParser/JPetCmdParser.cpp
```diff
@@ -61,6 +65,18 @@ std::vector<JPetOptions> JPetCmdParser::parseAndGenerateOptions(int argc, const
...
+    JPetOptions optionsFromJson = JPetOptionsJson::createOptionsFromFile(variablesMap["json"].as<std::string>());
+    std::pair<std::map<char,int>::iterator,bool> ret;
+    for(auto const& iter : optionsFromJson.getOptions()){
```
wkrzemien: auto opts = optionsFromJson.getOptions();
for(auto const& iter : opts){
....
}
otherwise you are calling getOptions method every time

JPetCmdParser/JPetCmdParser.cpp
```diff
@@ -61,6 +65,18 @@ std::vector<JPetOptions> JPetCmdParser::parseAndGenerateOptions(int argc, const
...
+      ret = variablesMap.insert( std::pair<std::string, po::variable_value>(std::string(iter.first),po::variable_value(iter.second)) );
```
wkrzemien: I got lost a bit in this code

wkrzemien:
Just to redo it

PR #70, opened 2017-01-23
User options

wkrzemien:
This code is the adaptation of Klara's code to read additional user option from json file. 
It also includes the changes proposed in #69, so it should be merged after.
The user can add special flag to read in a json configuration file with a set of options (key:value) ,which are added to the options read from the command line and propagated to all the analysis modules.

### DIFF
--- a/JPetOptions/JPetOptionsTools.h
+++ b/JPetOptions/JPetOptionsTools.h
@@ -1,0 +1,26 @@
+/**
+ *  @copyright Copyright 2017 The J-PET Framework Authors. All rights reserved.
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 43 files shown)


JPetOptions/JPetOptionsTools.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef JPETOPTIONSTOOLS_H
+#define JPETOPTIONSTOOLS_H
+#include <map>
+
+namespace jpet_options_tools
+{
+typedef std::map<std::string, std::string> Options;
+bool createConfigFileFromOptions(const Options& options, const std::string& outFile);
+Options createOptionsFromConfFile(const std::string& inFile);
```
alekgajos: "createOptionsFromConfigFile" would be a name more consitstent with the other method.

wkrzemien: ok. Done

PR #71, opened 2017-01-25
Add printCurrentOptionsToLog that sends task current options sets to LOG

wkrzemien:


JPetTaskChainExecutor/JPetTaskChainExecutor.cpp
```diff
@@ -82,6 +83,14 @@ bool JPetTaskChainExecutor::process()
   return true;
 }
 
+void JPetTaskChainExecutor::printCurrentOptionsToLog(const JPetOptions::Options& currOpts)
+{
+  INFO("Current options:");
+  for (const auto & el : currOpts) {
+    INFO("key:" + el.first + " value:" + el.second);
```
alekgajos: This is just a suggestion. Since half of our users do not know what a map/dictionary is and probably are not familiar with the "key"-"value" naming convention, perhaps it would be clearer to write to the log something like:
el.first + " = " el.value

wkrzemien: Ok, I will change it as you suggest.

alekgajos:
Won't this print the options several times, once for each task in the chain?
Also in case of multiple input files (which can happen in principle) I think this code would print the options 
multiple times, at least once for each input file.

PR #72, opened 2017-01-30
Get tomb channel map

wkrzemien:
Add JPetParamBankTools with getTOMBMap method. Method returns a map with the relations between layer, slot, side, thresholds and TOMB channel number. Also, add some tests.
This method is needed to translate time calibration parameter.

I merged code from Alek and Oleksandr's approaches

Please review it @alekgajos 

@alexkernphysiker  please have a look, cause I had to change slightly your interface (AbstractLargeBarrelMapping is replaced by JPetGeomMappingInterface)


alekgajos:
I think the mapping is not correct in the present form. The numbers of layers and slots in the calibration file should be the ones from the real large barrel numbering, i.e. layers from 1 to 3 and slots 1-48/96 for each layer. Using layer.getID() and barrel_slot.getID() would return database ID-s which are not strictly related to those numbers. We sould have to use something like LargeBarrelMapping or its equivalent from Oleksandr.

Moreover, you should use getLocalChannelNumber() instead of getThreshold() which returns a real number with set threshold value.

JPetParamBankTools/JPetParamBankTools.cpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+    if (ch->getPM().isNullObject() ||
+        ch->getPM().getScin().isNullObject() ||
+        ch->getPM().getScin().getBarrelSlot().isNullObject() ||
+        ch->getPM().getScin().getBarrelSlot().getLayer().isNullObject()) {
+      ERROR("Some parameter object is not set!");
+      errorOccured = true;
+      break;
+    }
+    auto threshold = ch->getThreshold();
```
alekgajos: Use ch->getLocalChannelNumber() instead.

JPetParamBankTools/JPetParamBankTools.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto pm_side = ch->getPM().getSide();
+    auto layer_id = ch->getPM().getScin().getBarrelSlot().getLayer().getID();
+    auto barrel_slot_id = ch->getPM().getScin().getBarrelSlot().getID();
```
alekgajos: layer_id and barrel_slot_id must be passed to LargeBarrelMapping or something similar in order to be recalculated into the actual numbers from the real (not database) numbering.

alexkernphysiker:
I don't see any changes in the interface class except it's renaming. Am I right?

PR #74, opened 2017-02-06
Changes in Barrel Extensions needed for last version of time synchronization software

alexkernphysiker:


JPetLargeBarrelExtensions/BarrelExtensions.cpp
```diff
@@ -115,13 +115,13 @@ void TOT_Hists::init(const JPetTaskInterface::Options& opts){
 TOT_Hists::TOT_Hists(const char* name, const char* description)
 :LargeBarrelTask(name,description){}
 TOT_Hists::~TOT_Hists(){}
-void TOT_Hists::createTOTHistos(const std::string& suffix){
+void TOT_Hists::createTOTHistos(const std::string& suffix,const size_t bins,const double&min,const double&max){
```
wkrzemien: const double& makes little to no sens. It is unnnecessary optimalization to avoid
see e.g.:
http://stackoverflow.com/questions/14013139/is-it-counter-productive-to-pass-primitive-types-by-reference
and
http://josephmansfield.uk/articles/need-value-pass-by-value.html

alexkernphysiker:
If you don't want to see const reference, I'll remove it, though I still disagree. I can submit it because there's no long cycle or deep recursion. Nevertheless look at the following situation:

SomeClass::SomeClass(const double v):somefield(v){}

what do we have? we create one copy of the value transferred and name it v. Then we copy v to somefield.
Two copies. double if larger type than pointer.
Or we can have similar situation:
void SomeFunc3(const double v){...do smth with v..}
void SomeFunc2(const double v){
    ......
     SomeFunc3(v);
    ......
}
void SomeFunc1(const double v){
    ......
     SomeFunc2(v);
    ......
}
Here, if we call SomeFunc1, three copies of v are created instead of one. 
If we want just to transfer the parameters to another function it's not necessary to create more copies than one. 
Here the doubles are just transferred to TH1F's constructor that accepts doubles as values maybe because this class was created when there was no such stuff as const reference in language. I suppose that additional copies are also created there. 
That's why I consider following such kind of a general opinions about good or bad coding style without thinking about all details to be a rather brave solution.

PR #76, opened 2017-02-15
removing duplicated method of getting scin id in hit class

kkacprzak:
Fixing error #813 from redmine.


wkrzemien:
Thanks. I think you should also fix the JPetLorTest since this is used there.

kkacprzak:
OK, done that as well - changes in JPetLORTest.

PR #77, opened 2017-02-17
New method to create sinogram

grey277:
I added new method to create sinogram, its create little different sinogram then previous one(shifted by center), but it pass all tests(0 and 90 degree, for 45 degrees it gives nearby diagonal values).

### DIFF
--- a/modules/JPetRecoImageTools/JPetRecoImageTools.cpp
+++ b/modules/JPetRecoImageTools/JPetRecoImageTools.cpp
@@ -293,34 +293,163 @@ JPetRecoImageTools::Matrix2DProj JPetRecoImageTools::backProject(Matrix2DProj &s
   return reconstructedProjection;
 }
 
+void JPetRecoImageTools::None(Matrix2DProj &sinogram)
+{
+  int nAngles = sinogram[0].size();
+  int nScanSize = sinogram.size();
+  int pow = std::round(std::log(nScanSize) / std::log(2.0));
+  int padlen = std::round(std::pow(2.0, pow + 1));
+  std::vector<double> Filt(padlen);
+  int highest = std::round(padlen / 2);
+  Filt[0] = 0.;
+  for (int j = 0; j < highest; j++)
+  {
+    Filt[j] = 1. / (double)padlen;
+    Filt[padlen - j] = 1. / (double)padlen;
+  }
+  Filt[highest] = 1. / (double)padlen;
+
+  doFFT1D(sinogram, nAngles, nScanSize, padlen, Filt);
+}
+
 void JPetRecoImageTools::RamLak(Matrix2DProj &sinogram)
 {
-  int nang = sinogram[0].size();
-  int ny = sinogram.size();
-  int pow = std::round(std::log(ny) / std::log(2.0));
+  int nAngles = sinogram[0].size();
+  int nScanSize = sinogram.size();
+  int pow = std::round(std::log(nScanSize) / std::log(2.0));
   int padlen = std::round(std::pow(2.0, pow + 1));
   std::vector<double> Re(padlen);
   std::vector<double> Im(padlen);
-  std::vector<double> col(ny);
+  std::vector<double> col(nScanSize);
   std::vector<double> Filt(padlen);
-  int Highest = std::round(padlen / 2);
-  double nu;
+  int highest = std::round(padlen / 2);
+  double nu = 0.;
   Filt[0] = 0.;
-  for (int j = 0; j < Highest; j++)
+  for (int j = 1; j < highest; j++)
   {
     nu = (double)j / (double)padlen;
     Filt[j] = nu;
     Filt[padlen - j] = nu;
   }
-  Filt[Highest] = (double)Highest / (double)padlen;
-  for (int k = 0; k < nang; k++)
+  Filt[highest] = (double)highest / (double)padlen;
+
+  doFFT1D(sinogram, nAngles, nScanSize, padlen, Filt);
+}
+
+void JPetRecoImageTools::SheppLogan(Matrix2DProj &sinogram)
+{
+  int nAngles = sinogram[0].size();
+  int nScanSize = sinogram.size();
+  int pow = std::round(std::log(nScanSize) / std::log(2.0));
+  int padlen = std::round(std::pow(2.0, pow + 1));
+  std::vector<double> Re(padlen);
+  std::vector<double> Im(padlen);
+  std::vector<double> col(nScanSize);
+  std::vector<double> Filt(padlen);
+  int highest = std::round(padlen / 2);
+  double nu = 0.;
+  Filt[0] = 0.;
+  for (int j = 1; j < highest; j++)
   {
-    for (int l = 0; l < ny; l++)
+    nu = (double)j / (double)padlen;
+    Filt[j] = std::sin(M_PI * nu) / M_PI;
+    Filt[padlen - j] = Filt[j];
+  }
+  nu = (double)highest / (double)padlen;
+  Filt[highest] = std::sin(M_PI * nu) / M_PI;
+
+  doFFT1D(sinogram, nAngles, nScanSize, padlen, Filt);
+}
+
+void JPetRecoImageTools::Cosine(Matrix2DProj &sinogram)
+{
+  int nAngles = sinogram[0].size();
+  int nScanSize = sinogram.size();
+  int pow = std::round(std::log(nScanSize) / std::log(2.0));
+  int padlen = std::round(std::pow(2.0, pow + 1));
+  std::vector<double> Re(padlen);
+  std::vector<double> Im(padlen);
+  std::vector<double> col(nScanSize);
+  std::vector<double> Filt(padlen);
+  int highest = std::round(padlen / 2);
+  double nu = 0.;
+  Filt[0] = 0.;
+  for (int j = 1; j < highest; j++)
+  {
+    nu = (double)j / (double)padlen;
+    Filt[j] = nu * std::cos(M_PI * nu);
+    Filt[padlen - j] = Filt[j];
+  }
+  nu = (double)highest / (double)padlen;
+  Filt[highest] = nu * std::cos(M_PI * nu);
+
+  doFFT1D(sinogram, nAngles, nScanSize, padlen, Filt);
+}
+
+void JPetRecoImageTools::Hamming(Matrix2DProj &sinogram)
+{
+  int nAngles = sinogram[0].size();
+  int nScanSize = sinogram.size();
+  int pow = std::round(std::log(nScanSize) / std::log(2.0));
+  int padlen = std::round(std::pow(2.0, pow + 1));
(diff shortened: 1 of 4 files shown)


modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -94,10 +93,65 @@ JPetRecoImageTools::Matrix2DProj JPetRecoImageTools::sinogram(Matrix2D& emission
...
+}
+
+double JPetRecoImageTools::calculateProjection2(int step, double cos, double sin, 
+                                                int imageSize, double center, double length, 
+                                                std::function<double(int, int)> matrixGet)
+{
+  double stepMinusCenter = step - center;
+  double xtmp = center + stepMinusCenter * cos;
+  double ytmp = center - stepMinusCenter * sin;
+  //int nmin = std::floor(center - std::sqrt((length - (stepMinusCenter * stepMinusCenter))));
```
wkrzemien: remove commented code

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -94,10 +93,65 @@ JPetRecoImageTools::Matrix2DProj JPetRecoImageTools::sinogram(Matrix2D& emission
...
+                                                std::function<double(int, int)> matrixGet)
+{
+  double stepMinusCenter = step - center;
+  double xtmp = center + stepMinusCenter * cos;
+  double ytmp = center - stepMinusCenter * sin;
+  //int nmin = std::floor(center - std::sqrt((length - (stepMinusCenter * stepMinusCenter)))); 
+  //TODO Repair, when step is last step in emissionMatrix it's calculating wrong value
+  //for nmin = 0 we checking a lot of unnecessary values
+  int nmin = 0;
+  //std::cout << "center: " << center << " length: " << length << " m * m: " << stepMinusCenter * stepMinusCenter << "\n";
```
wkrzemien: the same

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -94,10 +93,65 @@ JPetRecoImageTools::Matrix2DProj JPetRecoImageTools::sinogram(Matrix2D& emission
...
+  //std::cout << "nmin: " << nmin << "\n";
+  double p = 0.0; 
+  for(int n = nmin; n < imageSize - nmin; n++)
+  {
+    double nMinusCenter = (double)n - center;
+    int x = std::floor((xtmp - nMinusCenter * sin) + 0.5);
+    int y = std::floor((ytmp - nMinusCenter * cos) + 0.5);
+    p += matrixGet(x, y);
+    //std::cout << "x: " << x << " y: " << y << " val: " << matrixGet(x, y) << "\n";
```
wkrzemien: and here

grey277:
I added backprojection and Ram-Lak filter, it is backprojecting image correctly, but code need a lot of refactoring.

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -26,78 +26,88 @@ JPetRecoImageTools::JPetRecoImageTools() {}
...
 {
   assert(emissionMatrix.size() > 0);
   assert(emissionMatrix.size() == emissionMatrix[0].size());
   assert(nViews > 0);
   assert(nScans > 0);
   assert(angleBeg < angleEnd);
-  assert(minCutoff < scaleFactor);
+  assert(rescaleMinCutoff < rescaleFactor);
 
   //create vector of size nViews, initialize it with vector of size nScans
```
wkrzemien: Minor thing: the comment is outdated and probably not needed. Remove it

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -26,78 +26,88 @@ JPetRecoImageTools::JPetRecoImageTools() {}
...
-  Matrix2DProj proj(nViews, std::vector<double>(nScans));
+  Matrix2DProj proj(nScans, std::vector<double>(nViews));
 
   float stepsize = (angleEnd - angleBeg) / nViews;
   assert(stepsize > 0); //maybe != 0 ?
```
wkrzemien: another minor thing: is this comment really necessary?

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -84,12 +109,47 @@ class JPetRecoImageTools
                                     int nScans,
                                     InterpolationFunc& interpolationFunction
                                    );
+  static Matrix2DProj sinogram2(Matrix2D& emissionMatrix,
```
wkrzemien: You should add some comments to those functions.
Also sinogram1 and singogram2 are not very good names - so please add at least some comments what are the main differences.

modules/JPetRecoImageTools/JPetRecoImageToolsTest.cpp
```diff
@@ -121,80 +121,115 @@ BOOST_AUTO_TEST_CASE(linear)
...
+/// 4 5 6
+/// 7 8 9
+JPetRecoImageTools::InterpolationFunc func = JPetRecoImageTools::nearestNeighbour;
+JPetRecoImageTools::RescaleFunc func2= JPetRecoImageTools::nonRescale;
+const int kNScans = 3;
+const int kNViews = 2;
+auto results = JPetRecoImageTools::sinogram(matrix,kNViews,kNScans, 0 , 180, func, func2);
+for (auto i: results)
+for (auto j: i)
+std::cout <<j <<std::endl;
```
wkrzemien: some BOOST_REQUIRE is missing?

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -28,6 +32,27 @@ class JPetRecoImageTools
...
+        , result_type
+    >::type sequence_type;
+    // sequence_type is now a Boost.MPL sequence in the style of
+    // mpl::vector<int, double, long> if the signature of the
+    // analyzed functor were int(double, long)
+
+    // We now build a function type out of the MPL sequence
+    typedef typename ft::function_type<sequence_type>::type function_type;
+
+    std::function<function_type> function = std::move(f);
```
wkrzemien: I think calling function  as function is a bit misleadning :). Maybe func or myFunc

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -28,6 +32,27 @@ class JPetRecoImageTools
...
+    function(sinogram, std::forward(args)...);
```
wkrzemien: calling std::function "function" is not the best idea IMHO. maybe func, myFunc or whatever.

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -84,12 +109,47 @@ class JPetRecoImageTools
...
+                                     );
+
+  static Matrix2DProj backProject(Matrix2DProj& sinogram, int angles, 
+                                  RescaleFunc rescaleFunc, 
+                                  int rescaleMinCutoff,
+                                  int rescaleFactor);
+
+  static void None(Matrix2DProj &sinogram);
+
+  static void RamLak(Matrix2DProj &sinogram);
```
wkrzemien: The same here,

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -84,12 +109,47 @@ class JPetRecoImageTools
...
+
+  static void SheppLogan(Matrix2DProj &sinogram);
+  
+  static void Cosine(Matrix2DProj &sinogram);
```
wkrzemien: Cosine is the filter? So maybe add Filter to all filter functions ?

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -84,12 +109,47 @@ class JPetRecoImageTools
...
 private:
   JPetRecoImageTools();
   ~JPetRecoImageTools();
   JPetRecoImageTools(const JPetRecoImageTools&) = delete;
   JPetRecoImageTools& operator=(const JPetRecoImageTools&) = delete;
 
+  static void doFFT1D(Matrix2DProj &sinogram, int nAngles, int nScanSize,
+                      int padlen, std::vector<double> &Filt);
+
+  static void doFFT1D(std::vector<double> &Re, std::vector<double> &Im, int size, int shift, bool fast);
```
wkrzemien: general question? Is int large enough  e.g. for size? shouldn't it be long long? I don't know, just asking

wkrzemien: what is shift?, 
fast is fastFFT ?

modules/JPetRecoImageTools/JPetRecoImageTools.h
```diff
@@ -84,12 +109,47 @@ class JPetRecoImageTools
...
+  static Matrix2DProj backProject(Matrix2DProj& sinogram, int angles, 
+                                  RescaleFunc rescaleFunc, 
+                                  int rescaleMinCutoff,
+                                  int rescaleFactor);
+
+  static void None(Matrix2DProj &sinogram);
+
+  static void RamLak(Matrix2DProj &sinogram);
+
+  static void SheppLogan(Matrix2DProj &sinogram);
```
wkrzemien: On the other hand SheppLogan is a kind of phantom not a filter, so we should just distinguish between those cases

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -162,15 +228,319 @@ void JPetRecoImageTools::rescale(Matrix2DProj& matrix, double minCutoff, double
...
+    Filt[j] = std::sqrt(nu);
+    Filt[padlen - j] = Filt[j];
+  }
+  nu = (double)highest / (double)padlen;
+  Filt[highest] = std::sqrt(nu);
+
+  doFFT1D(sinogram, nAngles, nScanSize, padlen, Filt);
+}
+
+void JPetRecoImageTools::doFFT1D(Matrix2DProj &sinogram, int nAngles, int nScanSize,
```
wkrzemien: You will need to add some explanation or pseudo-equation in the header file.
Also, since I am assuming that you got inspired by some existing code (?), you could also somewhere write that this code was inspired by ... and give the link, or it was written based on FFT described in?

Also, one could adapt for testing some known FFT implementations e.g. this is one of the fastest: http://www.fftw.org/

wkrzemien:
Do I understand correctly that this one can be closed because the new updated version is available?

grey277:
Yes, new PR contains all commits from this PR + Fourier transform is performed using FFTW library, but results from new PR currently are not as good as from this PR, I think problem in new method is wrong calculation of distance from DC component in Fourier space, I will try to fix it when I will have some time.

PR #78, opened 2017-02-17
User options

kmuzalewska:
Create JPetOptionsGenerator to generate options form command line, json and to add necessary default options.


wkrzemien:
There are two conflicts - probably you must do rebase

JPetOptionsGenerator/JPetOptionsGenerator.h
```diff
@@ -0,0 +1,100 @@
...
+  inline const std::vector<std::string>& getFileNames(const po::variables_map& variablesMap) const {
+    return variablesMap["file"].as< std::vector<std::string> >();
+  }
+  inline bool isCorrectFileType(const std::string& type) const {
+    if (type == "hld" || type == "root" || type == "scope" || type == "zip") {
+      return true;
+    }
+    return false;
+  }
+
```
wkrzemien: Towards more generic solution:
please add two methods:
bool isOptionSet(const po::variables_map& variablesMap, const std::string& option) const; 
std::string getOptionValue(const po::variables_map& variablesMap, const std::string& option) const;

Next you can replace some existing methods by calling the more generic ones, and remove those.
E.g. you can replace isOutputPath(map); by isOptionSet(map, "outputPath"); 

This will not solve all the problems but you can reduce the number of functions considerably.

build/tests/sinogram.ppm
```diff
@@ -0,0 +1,183 @@
+P2
+256 180
```
wkrzemien: This file probably should be removed?

PR #82, opened 2017-03-01
Geometry  Mapping class

wkrzemien:
Add JPetGeomMapping class with set of methods that expresses the dependencies between various detector elements e.g.between layer, slot, side, thresholds and TOMB channel number. This method is needed to translate time calibration parameter.
The code is based on @alekgajos and @alexkernphysiker code, merged together.

Please review it @alekgajos

@alexkernphysiker please have a look, cause I had to change slightly your interface (AbstractLargeBarrelMapping is replaced by JPetGeomMappingInterface)


alexkernphysiker:
I don't see any changes in the interface class except it's renaming. Am I right?I don't see any changes in the interface class except it's renaming. Am I right?
Everything seems to be OK.

PR #86, opened 2017-03-08
Add getBranch metod to JPetReader

grey277:
Add getBranch method to JPetReader for EventDisplay to determinate type of file selected by user. TTree is holding branch named after class.

### DIFF
--- a/JPetReader/JPetReader.h
+++ b/JPetReader/JPetReader.h
@@ -81,8 +81,12 @@ public:
     else return false;
   }
 
+  virtual TBranch* getBranch(const char* name) {
+    return fTree->GetBranch(name);
+  }
+
 protected:
-  virtual bool openFile(const char* filename);
+  virtual bool openFile(const char *filename);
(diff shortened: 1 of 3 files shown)


JPetReader/JPetReader.h
```diff
@@ -81,8 +81,12 @@ class JPetReader : private boost::noncopyable, public JPetReaderInterface
     else return false;
   }
 
+  virtual TBranch* getBranch(const char* name) {
```
wkrzemien: Is this method really needed? I don't like this method because it is "implementation-dependend", all other methods in this interface are general, so in principle if we decide to replace the the tree by  e.g. vector of events it should work, without changing the interface.
If you must have direct access to the Branch try to find another solution e.g. create a class that inherits from JPetReader or consider https://root.cern.ch/7-using-ttreereader

grey277:
Ok, I found work around with getObjectFromFile getting tree and then branch name, so I close pr.

PR #87, opened 2017-03-09
Change int to float in JPetRawSignal

grey277:
JPetSigCh getThreshold is returning float, when getTimesVsThresholdValue was creating map there was round from float to int. It was returning bad results(usually 1 value instead of 4). I also added 'u' to remove compiler warnings about comparing int/uint.


JPetRawSignal/JPetRawSignalTest.cpp
```diff
@@ -114,21 +114,26 @@ BOOST_AUTO_TEST_CASE(GetMapOfTimesVsThrValueTest) {
...
+  sigch4.setThreshold(400.1f);
+  sigch4.setThresholdNumber(5);
 
   signal.addPoint(sigch1);
   signal.addPoint(sigch2);
   signal.addPoint(sigch3);
+  signal.addPoint(sigch4);
 
-  std::map<int, double> map;
+  std::map<float, float> map;
```
wkrzemien: using of float as a key value is somehow dangerous, because of the problem with comparing floating values.
One proposition is to change the structure of this map to:
int thresholdNumber: std::pair<float, float>  -> Threshold Value, time value
So for

wkrzemien:
Closed to to some git confilicst will be reopened soon

PR #89, opened 2017-03-12
Factories

kmuzalewska:
fabric with tests for:
JPetTimeWindow, JPetSigCh, JPetHit, JPetEvent, JPetPM, JPetScin, JPetBarrelSlot, JPetFEB, JPetLayer


JPetBarrelSlot/JPetBarrelSlot.cpp
```diff
@@ -36,6 +36,17 @@ JPetBarrelSlot::JPetBarrelSlot(bool isNull) :
 {
   SetName("JPetBarrelSlot");
 }
+JPetBarrelSlot JPetBarrelSlot::makeBarrelSlot(JPetLayer& p_layer, int id, bool isActive, std::string name, float theta, int inFrameID)
+{
+  JPetBarrelSlot barrelSlotObject;
```
wkrzemien: It is better to use the most expanded constructor JPetBarrelSlot::JPetBarrelSlot(int id, bool isActive, std::string name, float theta, int inFrameID) and then just set the missing fields.

JPetEvent/JPetEvent.h
```diff
@@ -56,6 +56,7 @@ class JPetEvent : public TNamed
   void addEventType(JPetEventType type);
   bool isTypeOf(JPetEventType type) const; /// check if the event is of a given type. Event can belong to more than one category e.g. k2Gamma|kScattered.
   bool isOnlyTypeOf(JPetEventType type) const;/// check if the event is of only given type e.g. isOnlyTypeOf(k2Gamma) will return false if the type is k2Gamma|kScattered.
+  static JPetEvent makeEvent(const JPetHit& hit, JPetEventType type);
```
wkrzemien: Here probably factory is not needed cause the exisiting constructor is enough

JPetPhysSignal/JPetPhysSignal.cpp
```diff
@@ -29,7 +29,21 @@ JPetPhysSignal::JPetPhysSignal() :
 
 JPetPhysSignal::~JPetPhysSignal()
 { }
+JPetPhysSignal::JPetPhysSignal(bool isNull):
+  fIsNullObject(isNull)
```
wkrzemien: I think fIsNullObject and all the related staff (Dummy etc) should be defined in the JPetBaseSignal class

wkrzemien:
This PR is obsolete. The new one is #95

PR #96, opened 2017-03-29
JPetStatistics file storage change

alekgajos:
JPetStatistics is now written to the ROOT file as a TDirectory (optionally with an additional subdirectory) instead of THashTable. JPetWriter was extended with a method allowing to write the contents of any TCollection into a TDirectory in a TFile.

Now the histograms contained in ROOT files produced by the Framework can be correctly added with "hadd" from ROOT.

Moreover, the subdirectory option allows for writing contents of several JPetStatistics objects (e.g.) from different tasks into the same file, which can be used in a version without writing of intermediate files.

### DIFF
--- a/JPetWriter/JPetWriter.cpp
+++ b/JPetWriter/JPetWriter.cpp
@@ -63,3 +63,57 @@ void JPetWriter::writeHeader(TObject* header)
   fTree->GetUserInfo()->AddAt(header, JPetUserInfoStructure::kHeader);
 }
 
+void JPetWriter::writeCollection(const TCollection * col, const char* dirname, const char* subdirname){
+
+  TDirectory * current =  fFile->GetDirectory(dirname);
+
+  if(!current){
+    current = fFile->mkdir(dirname);
+  }
+
+  assert(current);
+  
+  // use a subdirectory if requested by user
+  if(!std::string(subdirname).empty()){
+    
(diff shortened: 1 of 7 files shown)


JPetWriter/JPetWriter.h
```diff
@@ -69,6 +69,7 @@ class JPetWriter : private boost::noncopyable
     return fFile->WriteTObject(obj, name);
   }
 
+  void writeCollection(const TCollection * hash, const char* dirname, const char* subdirname="");
```
wkrzemien: could you add some comment what this function does, e.g. it is not obvious from the definition that it will create directory and subdirectory if they dont exist

JPetWriter/JPetWriter.cpp
```diff
@@ -63,3 +63,57 @@ void JPetWriter::writeHeader(TObject* header)
...
+
+  TObject * obj;
+  while((obj = it->Next())){
+    obj->Write();
+  }
+
+  fFile->cd();
+
+}
+
```
wkrzemien: very minor thing those empty lines are not needed

JPetWriter/JPetWriter.h
alekgajos: I have added the Doxygen method documentation in the .cpp file.

PR #103, opened 2017-05-22
Generate md5 from file and download json from http server

grey277:


wkrzemien:
It is postponed, so I close it.

PR #104, opened 2017-05-24
Automatic generation of a version header file based on Git tag and hash of the current revision

alekgajos:
Now CMake queries git for last tag and current commit hash, and creates a file version.h from a version.h.in template. version.h Is added to .gitignore so as not to confuse Git. Inside the framework, JPetTaskIO uses the definitions from this header file to store the version and revision hash in the JPetTreeHeader written to a ROOT file.


JPetTreeHeader/JPetTreeHeader.h
```diff
@@ -84,6 +90,9 @@ class JPetTreeHeader: public TObject{
   std::string stringifyDictionary() const;
   std::string stringifyHistory() const;
 
+  std::string fFrameworkVersion;
```
wkrzemien: Those should be still initialized in constructors

.gitignore
```diff
@@ -33,3 +33,6 @@ MakeFile
 unitTestData/*
 Examples/*
 DBHandler/Config/*
+
+#files generated by cmake
+version.h
```
wkrzemien: end of line missing

PR #113, opened 2017-07-09
Change to FFTW library for FT in filtering sinogram

grey277:
Methods to create sinogram from data matrix, filter sinogram in frequency domain and backproject to image.


JPetCmdParser/JPetCmdParserTest.cpp
```diff
@@ -95,7 +95,7 @@ BOOST_AUTO_TEST_CASE( parsing_zip_file )
   auto options = parser.parseAndGenerateOptions(argc, const_cast<const char**>(argv));
   BOOST_REQUIRE_EQUAL(options.size(), 1);
   auto option = options.at(0);
-  BOOST_REQUIRE_EQUAL(std::string(option.getInputFile()), "unitTestData/JPetCommonToolsTest/goodZip.gz");
```
wkrzemien: I think the previous version is better ->

CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ find_package(Boost 1.50.0 REQUIRED COMPONENTS
 include_directories(${Boost_INCLUDE_DIRS})
 add_definitions(${Boost_DEFINITIONS})
 
+find_package(FFTW REQUIRED)
```
wkrzemien: Is this package included in most of the platforms (Linux) or one needs to install it?
If the installation is needed then we need to change also the instructions how to install

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -162,15 +238,161 @@ void JPetRecoImageTools::rescale(Matrix2DProj& matrix, double minCutoff, double
...
+    {
+      sinogram[y][x] = outDouble[x + y * nAngles] / size;
+    }
+  }
+
+  fftw_free(out);
+  fftw_destroy_plan(plan);
+  fftw_destroy_plan(invPlan);
+  fftw_cleanup();
+}
```
wkrzemien: endline missing

modules/JPetRecoImageTools/JPetRecoImageTools.cpp
```diff
@@ -162,15 +238,161 @@ void JPetRecoImageTools::rescale(Matrix2DProj& matrix, double minCutoff, double
...
+
+double JPetRecoImageTools::RidgeletFilter(double radius)
+{
+  return std::sqrt(radius);
+}
+
+// see http://www.fftw.org/doc/One_002dDimensional-DFTs-of-Real-Data.html
+// http://www.fftw.org/fftw3.pdf
+void JPetRecoImageTools::doFFTW(Matrix2DProj &sinogram, FilterFunction &filter)
```
wkrzemien: I think this should be refactored in such a way, that you have some general function 
doFFT()
and inside you can call doFFTW()
that would allow you to change the implementation if at some point we decide to use another FFT implementation
I would even keep your previous version of FFT as an option

JPetCmdParser/JPetCmdParserTest.cpp
grey277: Yes, I think I forgot to pull some commits from master

grey277:
I don't think fftw3 is installed by default in ubuntu, I added info to INSTALL file in requirements section.
Also I fixed calculating position in Fourier space and now result is compared with old method.
I didn't refactor  doFFTW call, I started doing that, but I occurred weird linking error, I think it can be done in another PR.

wkrzemien:
@grey277  Its fine for me. Two more things:
1.Change the general PR description since you fixed the bug
2.you must modify .travis.yml to install fftw and better run all tests on your github account

PR #117, opened 2017-07-23
Using ParamGetterAscii to load parameters with ScopeLoader

timorl:
Removed ScopeParamGetter and now using ParamGetterAscii in its place. More info in commit messages and the mail I sent to Wojtek and Szymon.


alekgajos:
I only have one remark:

During execution of ScopeLoader-based program, the following is errors are logged:

> Error initialize():No TRBs in run 1
> Error initialize():No FEBs in run 1
> Error initialize():No frames in run 1
> Error initialize():No layers in run 1
> Error initialize():No TOMBChannels in run 1

As this will always be the case with Scope data and in this particular case this is not a reason to alert, perhaps it would be possible to check the "-t" option for "scope" and in that case to allow for missing TRBs, FEBs, frames, layers and TOMBChannels without signalling an error?

timorl:
Last commit should fix this by allowing the user to tell ParamManager which parameter objects not to load.

PR #106, opened 2017-05-31
Options

kmuzalewska:

PR #120, opened 2017-08-11
Options

wkrzemien:
New module options
Description will be provided soon.
This is Klara's work reforked to my account. Should be identical with
#106 (this one should be kept since all the comments are there).

Things to fix:
1)Some validators output messages with cerr, and some others do not.
2)init argument for users tasks should be some generalized option container
3)Is there no way to avoid using the type suffixes in the names of the command-line parameters? They are ok in case of the config file, but make the command line parameters quite awkward if one would use full parameter names. Also the help message is a bit confusing. Alternatively, perhaps there is a way in boost::program_options to disable long parameter names completely?


wkrzemien:
Closed because superseded by #123

PR #122, opened 2017-09-15
Extract filters to classes, add another method to calculate FT

grey277:
As mentioned in #113 I refactored FilterSinogram to call Fourier Transform algorithm with given filter and sinogram, also I extracted filters as classes that implements filter interface. FFTW and FFTSLOW have similar results, but FFTW is a bit better, also FFTW should be much faster(I didn't test it).


grey277:
Fixed commit history in #124

PR #145, opened 2017-11-30
Add template to get any type of objects

grey277:
Add template to get any type of histogram

### DIFF
--- a/Core/JPetStatistics/JPetStatistics.h
+++ b/Core/JPetStatistics/JPetStatistics.h
@@ -49,7 +49,10 @@ public:
   template <typename T>
   T& getHisto(const char* name)
   {
-    return dynamic_cast<T&>(*(fStats.FindObject(name)));
+    TObject* tmp = fStats.FindObject(name);
+    if (tmp == null)
+      ERROR("getHisto of " + name + " returned nullptr");
+    return dynamic_cast<T&>(*(tmp));
   }
 
   const THashTable* getStatsTable() const;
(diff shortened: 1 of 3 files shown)


Core/JPetStatistics/JPetStatistics.h
```diff
@@ -46,6 +46,12 @@ class JPetStatistics: public TObject
   void createCounter(const char* name);
   double& getCounter(const char* name);
 
+  template <typename T>
+  T& getHisto(const char* name)
+  {
```
wkrzemien: Please add ERROR printing in case if FindObject returns nullptr.

Core/JPetStatistics/JPetStatistics.h
```diff
@@ -46,6 +46,12 @@ class JPetStatistics: public TObject
   void createCounter(const char* name);
   double& getCounter(const char* name);
 
+  template <typename T>
+  T& getHisto(const char* name)
```
wkrzemien: I think this should be called getObject cause it's more general.

Could you also replace  all implemnetations of other function by calling your function .
e.g. createHistogram(...) {
 return createObject(...);
}
etc.
createCanvas(...) {
 return createObject(..);
}
The code is definitely redundant

grey277:
I do not think inlining any of this methods are necessary, if I am not wrong, all in class calls are inlined automated and also nearly all template calls are inlined too(if they are not specialized), if I am wrong, I can inline them.

PR #146, opened 2017-12-06
CMakeList changed to be compatible with ROOT 5 and 6

Alvarness:
If Root version is greater than 6.0 than cmake prints information in red that this version is not recomended yet. However project still builds and compiles.


cmake/FindROOT.cmake
```diff
@@ -1,19 +1,23 @@
...
 # This module sets up ROOT information
 # It defines:
-# ROOT_FOUND          If the ROOT is found
-# ROOT_INCLUDE_DIR    PATH to the include directory
-# ROOT_INCLUDE_DIRS   PATH to the include directories (not cached)
-# ROOT_LIBRARIES      Most common libraries
-# ROOT_<name>_LIBRARY Full path to the library <name>
-# ROOT_LIBRARY_DIR    PATH to the library directory
-# ROOT_DEFINITIONS    Compiler definitions and flags
+# ROOT_FOUND             If the ROOT is found
```
wkrzemien: I was thinking that you want to remove the fallback catalogue completely, but it seems that you just moving its content. Why do you need FindROOT.cmake at all? This file is version-specific
Is it not included in the standard root installation directory? I think the idea of this fallback,  was to use "our" fallback version only if you cannot find it in your default installation. Now it will be used by default?

CMakeLists.txt
```diff
@@ -130,6 +121,14 @@ generate_root_dictionaries(DICTIONARIES SOURCES ${SOURCES}
   INCLUDE_DIRS ${FOLDERS} #pass variable with include dirs for generating dictionaries
   )
 
+if(ROOT_VERSION VERSION_GREATER 6.0)
+  # forech(pmc ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/Dictionaries)
```
wkrzemien: remove this line, pls

PR #147, opened 2017-12-12
DB Handler purge, farewell libpqxx

kkacprzak:
- removing of old database connection tools
- no libdbhandler.so build, changes in CMakeLists.txt
- removing dependency from libpqx
- necessary changes in ParamManager and ParamGetter
-  changes in  INSTALL instructions


ParametersTools/JPetParamGetter/JPetParamGetter.h
```diff
@@ -33,6 +33,8 @@ typedef std::map<int, int> ParamRelationalData;
 class JPetParamGetter
 {
 public:
+
+  JPetParamGetter();
```
grey277: JPetParamGetter is an abstract class and should not have any constructor.

ParametersTools/JPetParamManager/JPetParamManager.h
```diff
@@ -45,14 +43,11 @@ class JPetParamManager
...
+  JPetParamManager() : fParamGetter(), fBank(0), fIsNullObject(false) {}
+  JPetParamManager(bool isNull) : fParamGetter(), fBank(0), fIsNullObject(isNull) {}
   JPetParamManager(JPetParamGetter* paramGetter) : fParamGetter(paramGetter), fBank(0) , fIsNullObject(false) {}
   JPetParamManager(JPetParamGetter* paramGetter, std::set<ParamObjectType> expectMissing) : fParamGetter(paramGetter), fExpectMissing(expectMissing), fBank(0) , fIsNullObject(false) {}
-  /// Special constructor to create NullObject.
-  /// This object can be returned if JPetParamManager is not created,
-  /// and the const& is expected to be returned.
-  explicit JPetParamManager(bool isNull) : fParamGetter(new JPetDBParamGetter()), fBank(0), fIsNullObject(isNull) {}
-  ~JPetParamManager();
+  virtual ~JPetParamManager();
```
grey277: You only need virtual destructor when at least one of class methods is virtual.

ParametersTools/JPetParamManager/JPetParamManager.h
```diff
@@ -45,14 +43,11 @@ class JPetParamManager
   /// factory method to produce JPetParamManager instance based on provided options
   static std::shared_ptr<JPetParamManager> generateParamManager(const std::map<std::string, boost::any>& options);
 
-  JPetParamManager() : fParamGetter(new JPetDBParamGetter()), fBank(0), fIsNullObject(false) {}
+  JPetParamManager() : fParamGetter(), fBank(0), fIsNullObject(false) {}
```
grey277: I think JPetParamManager do not need constructor without any arguments, it is not used anywhere and it resolves problem with no argument JPetParamGetter constructor

ParametersTools/JPetParamGetter/JPetParamGetter.h
kkacprzak: removed

ParametersTools/JPetParamManager/JPetParamManager.h
kkacprzak: removed 'virtual', but destructor is needed

ParametersTools/JPetParamManager/JPetParamManager.h
kkacprzak: no argument constructor is needed in JPetParamManger.cpp line 41
Constructors JPetParamManager()  and JPetParamManager(bool isNull) are needed for cases when no local db file was provided. I removed fParamGetter() from them, then it is enough.
But is it ok?

kkacprzak:
Small fix in example VelocityCalibration requested.

grey277:
It is still not valid solution, all fields should be initialized in constructor and now fParamGetter is not(also tests are not passing). I think best solution here will be creating new class, JPetParamGetterDummy, that will inheritance from JPetParamGetter and implements all virtual methods to return empty maps. It will allow you to initialize fParamGetter from JPetParamManager with that class if JPetParamManager constructor do not take JPetParamGetter as argument.

kkacprzak:
Together with Wojtek we talked it over and appears that it was sufficient just to remove all DB handler references and initialize some JPetParamManagers fields as null pointers.

wkrzemien:
After accepting another PR, some conflicts appeared. Please remove them, so I could check your PR and accept if possible,

kkacprzak:
Resolved

PR #148, opened 2017-12-12
Change unzip and unpack to first task in JPetManager

grey277:
Also fix to correctly change next module extension after unzipping


wkrzemien:
Can you check why tests are not passing?

grey277:
The problem was const char* from options that was passed to const std::string& in JPetUnpaker, it was copied into local field in class, but as pointer not copy of value.

PR #149, opened 2017-12-13
Changed CMakeLists to work with ROOT 5 and 6

Alvarness:
Travis is using repositories to install root-system and not setting enviroment variables. Therefore a little workaround was needed - proper cmake file is downloaded.

However user does not need to download that file - after running script bin/thisroot.[c]sh every ROOT connected enviroment variable is set and cmake will find apropriate files on it's own


.travis.yml
```diff
@@ -8,14 +8,15 @@ os:
...
-  - sudo apt-get install -y cmake 
+  - sudo apt-get install -y libboost-all-dev
+  - sudo apt-get install -y libpqxx-3.1 libpqxx3-dev
+  - sudo apt-get install -y cmake
   - sudo apt-get install -y libtinyxml2-dev
   - sudo apt-get install -y doxygen
   - sudo apt-get install -y root-system
   - sudo apt-get install -y libfftw3-3 libfftw3-dev
 script:
+    - wget https://raw.githubusercontent.com/root-project/root/v5-34-00-patches/etc/cmake/FindROOT.cmake -P $PWD/cmake
```
wkrzemien: Does it work also for other versions of ROOT v5 ? or only  v5-34?

CMakeLists.txt
```diff
@@ -7,7 +7,8 @@
 
 cmake_minimum_required(VERSION 2.6)
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake $ENV{ROOTSYS})
+# list(APPEND CMAKE_PREFIX_PATH )
```
wkrzemien: Pls, remove commented line

Alvarness: Done (it was some code-leftover). Also removed adding environment variable - on Travis it is not set either way

.travis.yml
Alvarness: It should work on all versions of ROOT above 5.32 (reason why we had fallback procedure).

This only relevant on Travis (or when ROOT is installed from repositories) - there is no file FindROOT.cmake on a whole computer (no developer files/ apt-get install root-system-dev package). Package 5.34 was chosen because it is version of ROOT available in repositories.

Alvarness:
Every time I was running framework it was with ROOT6 - even with previous releases (for example at the time of writing HitFinder module). And Unpacker worked then - however I have not made any tests related to his pull request.

As for the Travis tests - that would require either downloading approximately 100MB precompiled binary from ROOT web page and instaling it or making our own .deb package and then downloading it. Both ways seems doable.

wkrzemien:
I think downloading from the official site would be ok. 

Can you try to use this version with ROOT6 and check that everything is working according to this instruction:
http://sphinx.if.uj.edu.pl/~alek/framework_integration_tests/framework_integration_tests.txt

+ run all tests and check that there are no errors.

Alvarness:
Unfortunately there seems to be error with unpacker:
"Error exec():The hld file doesnt exist: àŸi"
Log in the attachment
[JPet_a8363fa5-8375-492a-b354-7bcc7367098d.log](https://github.com/JPETTomography/j-pet-framework/files/1567191/JPet_a8363fa5-8375-492a-b354-7bcc7367098d.log)

grey277:
Could you please upload your newest version and resolve conflict? Also please update framework to newest developer version(with jenkins folder) so jenkins could properly test is both root 5 with gcc 4.9 and 6 with gcc 5.4 is compilable with this PR

Alvarness:
This version is working on my PC with
ROOT 6.08/06
g++ 5.4
gcc 
It also builds .pcm dictionaries in proper directory
Tested with Alek's integration tests - I need access to jenkins to fix that checks

grey277:
All tests are defined in jenkins/root_6/docker-entrypoint.sh file, so to change tests, you do not need direct access to Jenkins. 
Line: "rc=$?; if [[ $rc != 0 ]]; then exit $rc; fi" is just for returning error code if some statement fails, and should be added after every command(later I will wrap it into function).
All environment are defined in Dockerfile.

For now it seems that jenkins can't find root and trying to use fallback, but it is not here:
CMake Error at CMakeLists.txt:20 (find_package):
  By not providing "FindROOT.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "ROOT", but
  CMake did not find one.

  Could not find a package configuration file provided by "ROOT" with any of
  the following names:

    ROOTConfig.cmake
    root-config.cmake

  Add the installation prefix of "ROOT" to CMAKE_PREFIX_PATH or set
  "ROOT_DIR" to a directory containing one of the above files.  If "ROOT"
  provides a separate development package or SDK, be sure it has been
  installed.

Do you have some special environment variables set?

Alvarness:
Only thing I can see right now is that test failed - don't know why or where.

As for the environment variables - during compilation of ROOT user is requested to execute script "thisroot.sh" (and add it to bash.rc). This script sets "CMAKE_PREFIX_PATH" variable to ROOT directory and therefore to proper file FindROOT.cmake. That file differs between different ROOT versions so it should be provided by ROOT version installed on users PC.

grey277:
Ok, sorry, I forget I didn't set up access for your jenkins account. It is done, can you check, do you can view jobs/console output in jenkins?
I'm executing "source /root-system/bin/thisroot.sh"  before "cmake ..".

Alvarness:
Yes, I can see console log now. 

From what I see now root is build via ./configure rather than CMake - maybe that is causing problem (I'm compiling ROOT in the way it is build on Jenkins to see if file exists, update when it is finished)

Alvarness:
After compiling and installing ROOT with the same procedure as in Jenkins (and in the same location) CMAKE_PREFIX_PATH is set to "/root-system". And I'm getting the same error.

However, ROOT version installed by CMake, in location: "/home/user/root_install_dir" has CMAKE_PREFIX_PATH set to that location. And compilation runs without problems.

That suggests that CMake is not able to access/find file in location "/root-system". I have no idea why.

PR #153, opened 2018-01-02
add -t/-f option check

novoqoorsky:
Created a n-options wrapper class for this particular task and future options checks.


Options/JPetOptionValidator/JPetOptionValidator.cpp
```diff
@@ -16,7 +16,6 @@
 #include "./JPetOptionValidator.h"
 #include "./JPetCommonTools/JPetCommonTools.h"
 #include "./JPetLoggerInclude.h"
-#include "./JPetAdditionalValidators.h"
```
wkrzemien: I haved it to develop (we always make  PRs for develop). This probably is the reason why this line is deleted.
Can you put it back?

Options/JPetOptionValidator/JPetOptionValidator.cpp
```diff
@@ -87,6 +87,21 @@ bool JPetOptionValidator::isCorrectFileType(std::pair <std::string, boost::any>
...
 }
 
+bool JPetOptionValidator::isFileTypeMatchingExtensions(std::pair<std::string, boost::any> option)
+{
+  ManyOptionsWrapper optionsWrapper = any_cast<ManyOptionsWrapper>(option.second);
+  std::vector<boost::any> optionsVector = any_cast<std::vector<boost::any>>(optionsWrapper.getOptionsVector());
+  std::string fileType = any_cast<std::string>(optionsVector[0]);
+  std::vector<std::string> fileNames = any_cast<std::vector<std::string>>(optionsVector[1]);
+  for (const std::string &fileName : fileNames) {
+    if (fileName.substr(fileName.find_last_of('.') + 1) != fileType) {
```
wkrzemien: instead of fileName.substr ... you can use JPetCommonTools::extractFilenameSuffix(fileName)

Options/JPetOptionValidator/JPetOptionValidator.cpp
novoqoorsky: I deleted it on purpose as it was unused. Do you want me to put it back?

wkrzemien: no, no. Then it is good. If it is not used anywhere you can delete also the  files itselfs

Options/JPetOptionValidator/JPetOptionValidator.cpp
novoqoorsky: A little typo in the function name ;)

novoqoorsky:
You mean that the function is causing a shutdown? If so, that's probably because no other parts know about existence of the wrapper I've added. The function will work only if used like in JPetOptionValidatorTest - it's needed to explicitly pass a ManyOptionsWrapper object as a parameter.

Honestly, I don't really know where exactly the function is called, but wherever it is, an instance of ManyOptionsWrapper has to be created there.

grey277:
I mean validation didn't worked, -t/-f params was different, but there was no error while parsing cmd arguments, but it started first task(unzip because -t param was zip). 

I think validation is called from:
bool JPetOptionValidator::areCorrectOptions(const std::map<std::string, boost::any>& optionsMap, std::vector<std::string>& isOption)

but I do not know how/is it possible to extract -t param from there, and if you anyway need to extract -t param is it not better to do it in validation function? Then i think you won't need ManyOptionsWrapper.

novoqoorsky:
Seems like everything works fine now. Could you please verify?

PR #160, opened 2018-03-05
[WIP] Task looper

wkrzemien:
Attempt for the iterative task scheme.

Still some unit test examples must be added


Core/JPetTaskLooper/JPetTaskLooper.cpp
```diff
@@ -0,0 +1,62 @@
+/**
+ *  @copyright Copyright 2017 The J-PET Framework Authors. All rights reserved.
```
kkacprzak: (not that important) maybe 2018?

Core/JPetTaskLooper/JPetTaskLooper.cpp
```diff
@@ -0,0 +1,62 @@
...
+  for (auto subTask : subTasks) {
+    if (!subTask) {
+      ERROR("some subTasks are nullptr");
+      return false;
+    }
+  }
+  while (fIsCondition(inParams)) {
+    for (auto subTask : subTasks) {
+      subTask->init(inParams);
+      subTask->run(nullDataObject);
```
kkacprzak: I do not quite get it, why sub tasks run on null data?

Core/JPetTaskLooper/JPetTaskLooper.cpp
wkrzemien: Ok. Good catch

Core/JPetTaskLooper/JPetTaskLooper.cpp
wkrzemien: nullDataObject, cause I must pass something - and I didnt want to pass anything special

I will try to add this example + some description also in the code - it is true that it is hard to understand how it suppose to work

kkacprzak: Well, it is intended to be universal, so must be a bit abstract. So please ad example/test and description/documentation for doxygen.

wkrzemien:
I messed up with the git repo, so I must clean it up and reopen it later

PR #170, opened 2018-05-06
Update Dockerfile to use cmake instead of configure

grey277:
It is replacement of #149


alekgajos:
I will soon try to modify the CMake configuration to work with both ROOT 5 (built with autotools) and 6 (built with cmake). Eventually, we probably cannot remove all of the CMakeLists.txt part related to ROOT5, therefore I think we should wait with merging this PR until we have CMake working with both ROOT versions.

wkrzemien:
I have just accepted the ROOT6 PR. What should be done with this one?

alekgajos:
It should now be rebased onto develop. If Jenkins tests with ROOT 6 pass after the rebase, we will merge it.

grey277:
I rebased it onto develop. Now it is the question, do we want to test with root installed by autotools or by CMake? Also i can create 2 separate jenkins jobs to test 2 different root installations

PR #172, opened 2018-05-24
Universal CMake configuration

alekgajos:
The new CMake configuration should correctly find ROOT CMake modules independently of whether ROOT was build with autotools or with CMake. The fallback ROOTConfig.cmake module included in the framework should now be used only if no ROOTSYS nor CMake-related env variables are present (i.e. practically only in case of ROOT installed from a system repository).

Besides CMake changes, compilation with g++ 8 required adding includes for std::vector in a few places where it is used and exporting the "-pthread" linker flag to parent scope to correctly build the examples.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,6 +109,7 @@ foreach(dir ${FOLDERS_WITH_SOURCE})
 endforeach()
 
 list(APPEND FOLDERS ${CMAKE_CURRENT_SOURCE_DIR}) # for include to dictionaries for unpacker2
+list(APPEND FOLDERS ${Boost_INCLUDE_DIRS}) # add boost includes to dictionaries
 
 ##download test files
 # the script shouldn't do anything if the data is present and correct
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,30 +17,32 @@ if(NOT MSVC)
   add_definitions(-std=c++11 -Wall -Wunused-parameter)
 endif()
 
-foreach(mode QUIET REQUIRED)
-  find_package(ROOT 5 ${mode} COMPONENTS
-    Hist
-    Physics
-    RIO
-    Thread
-    Tree
-    )
-  if(ROOT_USE_FILE)
-    include(${ROOT_USE_FILE})
-  endif()
-  if(ROOT_FOUND)
-    break()
-  endif()
+################################################################################
+# Handling of ROOT 
+################################################################################
+
+# first, try the old-school ROOTSYS variable
+set(root_prefix $ENV{ROOTSYS})
+if(root_prefix)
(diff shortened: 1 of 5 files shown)


alekgajos:
There is not much to review, but please try to test if it builds and if tests pass on your systems. The more configurations we try, the better.

I have tested:
ROOT 5.34/26 (autotools), g++ 4.8.4
ROOT 6.12/06 (cmake), g++ 8.1.0
Our Travis has ROOT installed from system repository, which also  works

@grey277 The Jenkins test with ROOT 6 is failing, apparently due to some boost-related compilation error. I  will create a Redmine issue for that, please have a look at that error.

grey277:
I think I found the problem, it is necessary to add boost includes to dictionaries also, after:
list(APPEND FOLDERS ${CMAKE_CURRENT_SOURCE_DIR}) # for include to dictionaries for unpacker2
please add:
list(APPEND FOLDERS ${Boost_INCLUDE_DIRS}) # add boost includes to dictionaries

It should resolve the problem

alekgajos:
You were right, that was it exactly. Thanks!

I have also tested a configuration with one more g++ version:
ROOT 6.12/06 (cmake), 7.3.0

grey277:
Jenkins is also testing with default g++ version of ubuntu 16.1: g++ 5.3.1 so it is working with this also. Jenkins now is also installing root v6 with autotools instead of CMake so it also works

CMakeLists.txt
```diff
@@ -11,37 +11,41 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
...
+if(NOT ROOT_FOUND)
   # if we failed with CMake based find_package, use fallback
   # root-config version from cmake/fallback/
   list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/fallback)
-endforeach()
+  message(WARNING "No CMake modules from a ROOT installation found. Trying fallback mode.")
+endif()
 
-message(STATUS "ROOT version: ${ROOT_VERSION}")
+find_package(ROOT ${REQUIRED})
```
grey277: If ROOT package was found before in QUIET, does try to find it again won't cause problems?

alekgajos:
I don't think this is dangerous (the old version with a foreach loop did the same thing in fact) but I had included the second find_package with REQUIRED option mostly to have an automatic message about ROOT being found printed in every case. I have now moved the second call inside if as you suggested and added a custom message.

Also added one more way of finding ROOT's CMake modules (idea borrowed from Unpacker).

BlurredChoise:
It works very well for cmake 3.10 and ROOT 6.15
Great job :)

PR #173, opened 2018-05-31
MC wrapper

daria137:
- added GeantParser containing structures necessary for processing MC information
- main function JPetGeantParser; rest are structures inherited from Geant simulations
- included basic tests


alekgajos:
This PR was replaced by #175 for easier maintenance.

PR #176, opened 2018-07-01
Add function to calculate TOT

grey277:
Adding function to calculate TOT to JPetHit

### DIFF
--- a/DataObjects/JPetHit/JPetHit.cpp
+++ b/DataObjects/JPetHit/JPetHit.cpp
@@ -20,18 +20,16 @@
 
 ClassImp(JPetHit);
 
-JPetHit::JPetHit() :
-  TObject()
+JPetHit::JPetHit() : TObject()
 {
 }
 
-JPetHit::JPetHit(float e, float qe, float t, float qt, TVector3& pos, JPetPhysSignal& siga, JPetPhysSignal& sigb,
-                 JPetBarrelSlot& bslot, JPetScin& scin) :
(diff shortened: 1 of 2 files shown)


wkrzemien:
with the calculateTOT from EventCategorizersTools

kkacprzak:
This method does the same as EventCategorizerTools::calculateTOT(hit), with the use of maps instead of vectors.
So do we want it here? I think calculating the TOT should be a part of reconstruction procedures, not a general data structure.

grey277:
Generally yes, it should to the same.
One of the things with vector and sorting are missing thresholds, e.g if we will have from side A thr 1,2,3 and from side B 1,2,4 algorithm from EventCategorizerTools will calculate ToT over 3 thresholds and this algorithm will only include 2 first thresholds.
Beside of that on one of the framework meetings I think there was idea to add general method to calculate ToT to JPetHit object, but I could misunderstood something.

kkacprzak:
Still, I think the method should be specific to the analysis/reconstruction. I.e. in the current SignalFinder, there is no possibility to create signal with occupied thresholds 1,2,4, without 3rd one. TOT calculation can be done in different ways for different analysis approaches. In the end we want to use it to express somehow the energy of the hit, but way of calculating this value is up to reconstruction. If we want a general, agreed-on TOT calculation, then lets add a field of TOT to JPetHit structure and fill it in the reconstruction.

wkrzemien:
The function will be added to examples, so I am closing this PR

PR #177, opened 2018-07-03
TaskIO and JPetManager refactorization

wkrzemien:
That was supposed to be the tterative PR but meanwhile I did many refactorization staff, so I think it is better to split it into two PRs.
Changes in the backbone architecture:
0.Add make_unique implementation missing in C++11
1.Refactoring of JPetTaskIO:  separate JPetOutputHandler and JPetTaskIOTools. Also, add option that in case of empty outputFileType string passed to the constructor, the task will not write anything to the output.
2.Also, some changes of JPetScopeLoader
3. Create JPetTaskFactory and remove the Task creation logic from JPetManager
4.Add bunch of unique pointers.
5.Cleanup and simplification of JPetManager

JPetTaskIO code is still very dirty, however I think is better than the previous version


wkrzemien:
Superseeded by the PR #179

PR #178, opened 2018-07-05
Revision of documentation and comments

kkacprzak:
General documentation revision and comments clean-up, spelling corrections etc.
Doxygen generates TODO list, there are some clues what can be done next.

Redmine task [946](http://sphinx.if.uj.edu.pl/redmine/issues/946)


CMakeLists.txt
```diff
@@ -1,45 +1,46 @@
-# J-PET framework project main description file
-#
-# Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
-#
-# Description:
-# Builds all J-PET framework modules using CMake build generator.
+################################################################################
+## J-PET Framework Library
+## Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
```
JPETTomographyTester: Since we are changing every comments.
I would propose to remove it or rephrase it cause it suggests that Adam is the author of all the Framework :), which is clearly not true.

JPETTomographyTester:
By the way, we should also cross-check with the generated doxygen documentation

kkacprzak:
Doxygen documentation is good, I generated it and checked it as well. We can later upload it to sphinx. 
The problem why the tests do not pass is in several tests: JPetBarrelSlotTest, JPetManagerTest, JPetParamBankHandlerTaskTest, JPetReaderTest, JPetTaskChainExecutorTest. Those produce ROOTs segmentation violation while executing (all code builds without problem). I do not know why is that, because I changed only formatting to astyle.

CMakeLists.txt
kkacprzak: well I think he was the author of initial cmake script, that was modified later several times. Vestigial authorship

JPETTomographyTester:
Formatting cannot afect it, so there must be some changes. Could you try to go back to some previous commits and check the tests locally?

CMakeLists.txt
JPETTomographyTester: so maybe remove this Author line?

kkacprzak:
Corrections done.

wkrzemien:
Thx. Great work. I will revise it just to be sure and merge it later.

PR #182, opened 2018-07-14
Add option to print different levels of log 

grey277:
Add option to print different levels of log only or disable logging at all.


Core/JPetLoggerInclude.h
```diff
@@ -16,7 +16,7 @@
  *  Also general macro is defined that allows to create user own level of logging.
  */
 
- /**
+/**
   * @brief Configuration file for the Logger class
   * Four independent level of logging are defined: INFO, WARNING, ERROR and DEBUG.
```
wkrzemien: Check if the description is correct, and write the example how to use your macros

PR #183, opened 2018-07-16
 Add option to print different levels of log

grey277:
Add option to print different levels of log only or disable logging at all.
It is replacement of #182 because there was some problems with github and it didn't automatically update PR commits.

This PR is also removing uuid include and warnings: 
In file included from /framework-dependencies/include/boost/uuid/uuid.hpp:203:
/framework-dependencies/include/boost/uuid/detail/uuid_x86.hpp:42:5: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
    register __m128i mm = uuids::detail::load_unaligned_si128(data);
    ^~~~~~~~~

and changing log output file name to: JPet_%Y-%m-%d_%H-%M-%S.%N.log (%N means number of log) e.g.: JPet_2018-07-16_19-50-05.0.log

### DIFF
--- a/Core/JPetLoggerInclude.h
+++ b/Core/JPetLoggerInclude.h
@@ -47,10 +47,10 @@
 #define DEBUG(X) CUSTOM_LOG(JPetLogger::getInstance().getSeverity(), boost::log::trivial::debug, X)
 #define LOG(X, sev) CUSTOM_LOG(JPetLogger::getInstance().getSevarity(), sev, X)
 
-#define SET_MINIMAL_LOG_ERROR JPetLogger::getInstance().setLogLevel(boost::log::trivial::error)     // prints only error messages
-#define SET_MINIMAL_LOG_WARNING JPetLogger::getInstance().setLogLevel(boost::log::trivial::warning) // prints error + warning messages
-#define SET_MINIMAL_LOG_INFO JPetLogger::getInstance().setLogLevel(boost::log::trivial::info)       // prints error + warning + info messages
-#define SET_MINIMAL_LOG_DEBUG JPetLogger::getInstance().setLogLevel(boost::log::trivial::debug)     // prints error + warning + info + debug messages
+#define SET_MINIMAL_LOG_ERROR() JPetLogger::getInstance().setLogLevel(boost::log::trivial::error)     // prints only error messages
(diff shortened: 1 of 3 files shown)


Core/JPetLoggerInclude.h
```diff
@@ -30,37 +30,32 @@
...
-#define SET_MINIMAL_LOG_INFO setLogLevel(boost::log::trivial::info)       //prints error + warning + info messages
-#define SET_MINIMAL_LOG_DEBUG setLogLevel(boost::log::trivial::debug)     //prints error + warning + info + debug messages
-
-#define SET_MINIMAL_LOG_LEVEL(X) setLogLevel(X)
-
-//for backward compability
-#define ENABLE_DEBUG setLogLevel(boost::log::trivial::debug)
-#define DISABLE_DEBUG setLogLevel(boost::log::trivial::info)
+#define CUSTOM_LOG(logger, sev, X)                                                                                                                   \
+  if (true) {                                                                                                                                        \
```
BlurredChoise: Why here is this construction with **if(true){}else{(void)0}** ?? Is not better to remove this if{}else{} ?

Core/JPetLoggerInclude.h
```diff
@@ -30,37 +30,32 @@
...
+    BOOST_LOG_SEV(logger, sev) << boost::log::add_value("Line", __LINE__) << boost::log::add_value("File", __FILE__)                                 \
+                               << boost::log::add_value("Function", __func__) << X;                                                                  \
+  } else                                                                                                                                             \
+  (void)0
+
+#define INFO(X) CUSTOM_LOG(JPetLogger::getInstance().getSeverity(), boost::log::trivial::info, X)
```
BlurredChoise: Maybe more details when use INFO, ERROR, ..., LOG ? 
For me the most important is description when and how use LOG.

BlurredChoise:
_code review code - waiting for answers from PR aouthor_

Core/JPetLoggerInclude.h
grey277: It is for multi statements macros, for more information see: http://www.cs.technion.ac.il/users/yechiel/c++-faq/macros-with-multi-stmts.html

PR #184, opened 2018-07-20
Added double in OptionsTypeHandler

Alvarness:
Previously double wasn't allowed type

### DIFF
--- a/Options/JPetOptionsGenerator/JPetOptionsTypeHandler.cpp
+++ b/Options/JPetOptionsGenerator/JPetOptionsTypeHandler.cpp
@@ -20,7 +20,7 @@
 #include "./JPetLoggerInclude.h"
 namespace pt = boost::property_tree;
 
-const std::vector<std::string> JPetOptionsTypeHandler::kAllowedTypes = {"int", "std::string", "bool", "std::vector<std::string>", "std::vector<int>", "float"};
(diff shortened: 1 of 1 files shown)


DataObjects/JPetEvent/JPetEvent.cpp
```diff
@@ -44,7 +44,7 @@ void JPetEvent::addHit(const JPetHit& hit)
   fHits.push_back(hit);
 }
 
-std::vector<JPetHit> JPetEvent::getHits() const
+const std::vector<JPetHit>& JPetEvent::getHits() const
```
wkrzemien: Thx for the fix.
This file should not be modified, because it is not related to the problem

DataObjects/JPetEvent/JPetEvent.h
```diff
@@ -43,7 +43,7 @@ class JPetEvent : public TObject
   JPetEvent();
   JPetEvent(const std::vector<JPetHit>& hits, JPetEventType eventType = JPetEventType::kUnknown, bool orderedByTime = true);
 
-  std::vector<JPetHit> getHits() const;
+  const std::vector<JPetHit>& getHits() const;
```
wkrzemien: same here

wkrzemien:
Can you solve the conflicts?

PR #192, opened 2018-10-07
New data class for reconstructed o-Ps->3 events

alekgajos:
Necessary for analysis of o-Ps->3g in data of runs with extensive-size annihilation chambers.
The new class extends JPetEvent to contain the reconstructed annihilation point and time.

To be considered after #190.

### DIFF
--- a/DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
+++ b/DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
@@ -33,27 +33,33 @@ JPetOpsEvent::JPetOpsEvent(const std::vector<JPetHit>& hits, JPetEventType event
 }
 
 
-void JPetOpsEvent::Clear(Option_t *){
+void JPetOpsEvent::Clear(Option_t*)
+{
   fType = kUnknown;
   fHits.clear();
 }
 
-void JPetOpsEvent::setAnnihilationPoint(double x, double y, double z){
-  setAnnihilationPoint(TVector3(x,y,z));
+void JPetOpsEvent::setAnnihilationPoint(double x, double y, double z)
+{
+  setAnnihilationPoint(TVector3(x, y, z));
 }
 
-void JPetOpsEvent::setAnnihilationPoint(const TVector3& point){
+void JPetOpsEvent::setAnnihilationPoint(const TVector3& point)
+{
   fAnnihilationPoint = point;
 }
 
-const TVector3& JPetOpsEvent::getAnnihilationPoint() const{
+const TVector3& JPetOpsEvent::getAnnihilationPoint() const
+{
   return fAnnihilationPoint;
 }
 
-void JPetOpsEvent::setAnnihilationTime(double t){
+void JPetOpsEvent::setAnnihilationTime(double t)
+{
   fAnnihilationTime = t;
 }
 
-double JPetOpsEvent::getAnnihilationTime() const{
+double JPetOpsEvent::getAnnihilationTime() const
+{
   return fAnnihilationTime;
 }
--- a/DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
+++ b/DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
@@ -1,0 +1,59 @@
+/**
+ *  @copyright Copyright 2018 The J-PET Framework Authors. All rights reserved.
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may find a copy of the License in the LICENCE file.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  @file JPetOpsEvent.cpp
+ */
+
+#include "./JPetOpsEvent.h"
+
+ClassImp(JPetOpsEvent);
+
+JPetOpsEvent::JPetOpsEvent(): JPetEvent()
+{
+  /**/
+}
+
+JPetOpsEvent::JPetOpsEvent(const JPetEvent& event): JPetEvent(event)
+{
+  /**/
+}
+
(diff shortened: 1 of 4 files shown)


DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
```diff
@@ -0,0 +1,65 @@
...
+  /**/
+}
+
+JPetOpsEvent::JPetOpsEvent(const std::vector<JPetHit>& hits, JPetEventType eventType, bool orderedByTime):
+  JPetEvent(hits, eventType, orderedByTime)
+{
+}
+
+
+void JPetOpsEvent::Clear(Option_t*)
```
wkrzemien: Why capital C?

DataObjects/JPetOpsEvent/JPetOpsEvent.h
```diff
@@ -0,0 +1,52 @@
...
+  void setAnnihilationPoint(const TVector3& point);
+  void setAnnihilationTime(double t);
+
+  const TVector3& getAnnihilationPoint() const;
+  double getAnnihilationTime() const;
+
+protected:
+
+  TVector3 fAnnihilationPoint;
+  double fAnnihilationTime;
```
wkrzemien: set some default value to fAnnihilationTime;
Also one thing to consider, but it is up to you
maybe one could reuse TLorentVector to store both position and time ?

DataObjects/JPetOpsEvent/JPetOpsEvent.h
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Extends JPetEvent to contain the reconstructed annihilation point and time.
+ */
+
+class JPetOpsEvent : public JPetEvent
+{
+
+public:
+  JPetOpsEvent();
+  JPetOpsEvent(const JPetEvent& event);
```
wkrzemien: explicit JPetOpsEvent

DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
kkacprzak: Just for the record - in other classes we also use Clear functions with capital 'C', not in align with Framework convention. I thought that is an implementation of virtual void TObject::Clear(), is it?

kkacprzak:
Very useful next step of development, some design considerations:
- since this class aims in representing an event with physics meaning, can we make other directory for such classes? I.e. keep classes representing detector elements in ParamObjects, data with little physics interpretation in DataObjects and reconstructed physical events in PhysicalObjects or EventObjects? (proper name required)
- I guess in the procedure of reconstruction of time and position we get some uncertainty errors, so maybe add fields for that? Btw. we should think about error estimation in other procedures as well. 
- Maybe make it more general? So something like JPetPsDecay with enum like oPs, pPs or even oPs3g, oPs5g, pPs2g, pPs4g.

DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
wkrzemien: Ok. I had no idea. Where or how  do we use it.
Probably it should be also declared virtual then in case we add some subclasses later

kkacprzak: I am not sure, if we use Clear methods anywhere right now.

alekgajos: This is indeed and overload of TObject::Clear(). It is used implicitly in every class whose objects may be stored in JPetTimeWindow which internally uses TClonesArray.

alekgajos: You are probably right about making it virtual.
By the way, I got a little confused right now. If JPetEvent::Clear() was not virtual but overloaded a virtual method from TObject, the following would call Clear from JPetEvent of JPetOpsEvent?
```
JPetEvent * evt = new JPetOpsEvent();
evt->Clear();
```
?

DataObjects/JPetOpsEvent/JPetOpsEvent.h
alekgajos: I avoided TLorentzVector because its functionality is far beyond what is needed in this case, but in terms of data size it should be equivalent, so I'm not strictly against it. 
My only concern is that TLorentzVector may be overwhelming to some users - TVector3 is easier to explain.

alekgajos:
@kkacprzak, you are right about the need to store uncertainty as well.
Concerning the separate directory, I am not sure if these classes are distinct enough to separate them. 

The division of classes was based more on the technical reasons, i.e. DataObjects were everything that could be stored in JPetTimeWindows and, further, in our ROOT trees. This has a nice consequence that one can easily state that e.g. everyting from the single directory DataObjects must have a ClassDef and must implement the aforementioned Clear() method. If we split it, it would not be so straightforward. I propose to create an issue at Redmine to discuss it and in case we decide to do it, make a separate PR.

As for the generalization of this class, it is not a bad idea. @wkrzemien, what do you think?

DataObjects/JPetOpsEvent/JPetOpsEvent.cpp
wkrzemien: You are right although it is confusing for me. Maybe we could add override keyword, to make it more explicite that it is an overriden version.
From the standard:
"10.3 Virtual functions
2 If a virtual member function vf is declared in a class Base and in a class Derived, derived directly or indirectly from Base, a member function vf with the same name, parameter-type-list (8.3.5), cv-qualification, and ref-qualifier (or absence of same) as Base::vf is declared, then Derived::vf is also virtual (whether or not it is so declared) and it overrides Base::vf."

DataObjects/JPetOpsEvent/JPetOpsEvent.h
wkrzemien: Well, you can leave the interface as it is just change the implementatino to TLorentzVector and maybe add 2 methods to get/set the whole object.

wkrzemien: fAnnihilationTime = XXXX; if not :P

wkrzemien:
Agree on generalization

alekgajos:
This approach to adding custom and analysis-specific data classes was superseded by adding them to particular analysis directories in j-pet-framework-examples, see:
https://github.com/JPETTomography/j-pet-framework-examples/pull/138

PR #193, opened 2018-10-09
Unpacker-related user options

alekgajos:
* Adds possibility to define the path to TDC nonlinearity calibration file as a user parameter in the json file
* Allow to pass the path to TOT stretcher offsets calibration file as a user parameter alternatively to the "-c" command line parameter

### DIFF
--- a/Core/JPetUnpacker/JPetUnpacker.h
+++ b/Core/JPetUnpacker/JPetUnpacker.h
@@ -33,11 +33,14 @@ public:
   inline int getEventsToProcess() const { return fEventsToProcess; }
   inline std::string getHldFile() const { return fHldFile; }
   inline std::string getCfgFile() const { return fCfgFile; }
-  inline std::string getCalibFile() const { return fCalibFile; }
+  inline std::string getTOTCalibFile() const { return fTOTCalibFile; }
+  inline std::string getTDCCalibFile() const { return fTDCCalibFile; }
   void setParams(const std::string& hldFile,
                  int numOfEvents = 100000000,
                  const std::string& cfgFile = "conf_trb3.xml",
-                 const std::string& calibFile = "");
+                 const std::string& totCalibFile = "",
+		 const std::string& tdcCalibFile = ""
+		 );
 
   ClassDef(JPetUnpacker, 2);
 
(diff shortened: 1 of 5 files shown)


Core/JPetUnpacker/JPetUnpacker.h
```diff
@@ -30,23 +30,40 @@ class JPetUnpacker: public TObject
...
+                 const std::string& tdcCalibFile = ""
+                );
 
 private:
   Unpacker2* fUnpacker;
   int fEventsToProcess;
   std::string fHldFile;
   std::string fCfgFile;
-  std::string fCalibFile;
+  std::string fTOTCalibFile;
```
wkrzemien: I would just initialize all the variables here and remove the default constructor.
So, fUnpacker =nullptr; fEventsToProcess =-1; etc.
This is the recommended style since C++11

Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.h
```diff
@@ -29,12 +29,17 @@ class JPetUnzipAndUnpackTask: public JPetTask
   bool run(const JPetDataInterface& inData) override;
   bool terminate(JPetParams& outOptions) override;
   static void unpackFile(const std::string& filename, long long nevents,
-    const std::string& configfile, const std::string& calibfile);
+			 const std::string& configfile, const std::string& totCalibFile,
```
wkrzemien: small refactoring suggestion: since you have so many arguments  already it would be better to create some helper structure with all those fields and pass this one struct as argument.

alekgajos: I generally agree that this would be more elegant. However, in the view of the planned rewrite of the Unpacker (#971@Redmine), I think this is not worth the effort at this point.

Core/JPetUnpacker/JPetUnpacker.h
alekgajos: done

Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.h
wkrzemien: Ok. I agree

PR #195, opened 2018-10-28
Progress bar with current task name

kkacprzak:
This is a low-priority PR with adding a name of the task to progress bar, so it not only displays done percentage. It is useful when there are more than one task in chain, so current module name is visible.

### DIFF
--- a/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
+++ b/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
@@ -16,14 +16,7 @@
 #include "JPetProgressBarManager.h"
 #include <iostream>
 
-void JPetProgressBarManager::display(long long currentNumber, long long totalNumber) const
-{
-  std::cout << std::string(5, '\b');
-  std::cout << std::string(5, ' ');
-  std::cout << '\r' << getCurrentValue(currentNumber, totalNumber) << '%' << std::flush;
-}
-
-void JPetProgressBarManager::displayWithTaskName(
+void JPetProgressBarManager::display(
   std::string taskName, long long currentNumber, long long totalNumber
 ) const {
   std::cout << std::string(30, '\b');
--- a/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
+++ b/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
@@ -17,7 +17,7 @@
 #include <iostream>
 
 void JPetProgressBarManager::display(
-  std::string taskName, long long currentNumber, long long totalNumber
+  const std::string& taskName, long long currentNumber, long long totalNumber
 ) const {
   std::cout << std::string(30, '\b');
   std::cout << std::string(30, ' ');
--- a/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
+++ b/Core/JPetProgressBarManager/JPetProgressBarManager.cpp
@@ -14,11 +14,21 @@
  */
 
 #include "JPetProgressBarManager.h"
-#include <stdio.h>
+#include <iostream>
 
-void JPetProgressBarManager::display(long long done, long long end) const
+void JPetProgressBarManager::display(long long currentNumber, long long totalNumber) const
 {
-  printf("\r[%6.1f%%]", getCurrentValue(done, end));
+  std::cout << std::string(5, '\b');
+  std::cout << std::string(5, ' ');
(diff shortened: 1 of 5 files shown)


Core/JPetTaskIO/JPetTaskIO.h
```diff
@@ -49,47 +54,41 @@ struct TaskIOFileInfo {
...
   JPetParams getParams() const;
-  void displayProgressBar(int currentEventNumber, int numberOfEvents) const;
-
   bool isOutput() const;
   bool isInput() const;
 
 protected:
-  /// @return (isOK, inputFile, outputFileFullPath, isResetOutputPath) based on provided options. If isOK is set to false, that means that an error has occured.
-  virtual std::tuple<bool, std::string, std::string, bool> setInputAndOutputFile(const jpet_options_tools::OptsStrAny options) const;
+  virtual std::tuple<bool, std::string, std::string, bool> setInputAndOutputFile(
```
wkrzemien: why the doxygen comments are removed?

Core/JPetProgressBarManager/JPetProgressBarManager.cpp
```diff
@@ -14,11 +14,21 @@
  */
 
 #include "JPetProgressBarManager.h"
-#include <stdio.h>
+#include <iostream>
 
-void JPetProgressBarManager::display(long long done, long long end) const
+void JPetProgressBarManager::display(long long currentNumber, long long totalNumber) const
```
wkrzemien: is the old method used anywhere? Mabye just leave your new version and call it display and delete the old one.
Always, somebody can give an empty string.

Core/JPetProgressBarManager/JPetProgressBarManager.cpp
```diff
@@ -14,11 +14,21 @@
...
 {
-  printf("\r[%6.1f%%]", getCurrentValue(done, end));
+  std::cout << std::string(5, '\b');
+  std::cout << std::string(5, ' ');
+  std::cout << '\r' << getCurrentValue(currentNumber, totalNumber) << '%' << std::flush;
+}
+
+void JPetProgressBarManager::displayWithTaskName(
```
wkrzemien: I would change the name to display() and the name of the argument here as label, or string. In principle it can be any string

Core/JPetTaskIO/JPetTaskIO.h
kkacprzak: I moved it to .cpp file, I think it would be more useful in the file to see docs above the code, not only with declaration of the method in .h. For Doxygen is the same.

Core/JPetProgressBarManager/JPetProgressBarManager.cpp
kkacprzak: I am not sure, I checked but maybe I missed something. The compatibility probably is not super important in this case, so we can delete old method.

kkacprzak:
Requests applied

Core/JPetTaskIO/JPetTaskIO.cpp
```diff
@@ -295,31 +306,18 @@ const JPetParamBank& JPetTaskIO::getParamBank()
...
+JPetParamManager& JPetTaskIO::getParamManager()
 {
-  using namespace jpet_options_tools;
-  using namespace jpet_options_generator_tools;
-  auto oldOpts = oldParams.getOptions();
-  auto extraOpts = extraParams.getOptions();
-  const std::string stopIterationOptName = "stopIteration_bool"; /// @todo this is hardcoded and should be moved somewhere.
-  if (isOptionSet(extraOpts, stopIterationOptName)) {
-    oldOpts[stopIterationOptName] = getOptionValue(extraOpts, stopIterationOptName);
+  DEBUG("JPetTaskIO");
```
wkrzemien: why tthis code has been changed?

Core/JPetProgressBarManager/JPetProgressBarManager.h
```diff
@@ -16,13 +16,15 @@
...
+#include <string>
+
 /**
  * @brief Class managing the progress bar used in while processing events.
  */
 class JPetProgressBarManager
 {
 public:
-  void display(long long currentEventNumber, long long numberOfEvents) const;
+  void display(std::string taskName, long long currentEventNumber, long long numberOfEvents) const;
```
wkrzemien: std::string taskName ---> const std::string& taskName

kkacprzak: done

PR #197, opened 2018-11-16
Try to fix wrong range of events to process error

wkrzemien:
Change the logic in the JPetTaskIO, also add a bunch of tests in JPetInputHandler.
This is supposed to fix the problem with the event range, but I am not sure if this totally fix the problem.

### DIFF
--- a/Core/JPetTaskIO/JPetInputHandler.cpp
+++ b/Core/JPetTaskIO/JPetInputHandler.cpp
@@ -23,17 +23,6 @@ JPetInputHandler::JPetInputHandler()
   fReader = jpet_common_tools::make_unique<JPetReader>() ;
 }
 
-std::tuple<bool, long long, long long> JPetInputHandler::getEntryRange(const jpet_options_tools::OptsStrAny& options) const
-{
-  auto totalEntries = 0ll;
-  if (fReader) {
-    totalEntries = fReader->getNbOfAllEntries();
-  } else {
-    WARNING("no JPETReader,  totalEntries set to -1");
-    totalEntries = -1;
-  }
-  return JPetTaskIOTools::setUserLimits(options, totalEntries);
-}
 
 bool JPetInputHandler::openInput(const char* inputFilename, const JPetParams& params)
 {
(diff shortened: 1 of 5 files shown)


wkrzemien:
retest it please

wkrzemien:
I still must fix the remaing problem but if you have any more remarks, please let me know.

Core/JPetTaskIO/JPetTaskIO.cpp
```diff
@@ -113,7 +113,7 @@ bool JPetTaskIO::run(const JPetDataInterface&)
         if (isProgressBar(fParams.getOptions())) {
```
grey277: I have comment to line 95-97, if there was error in init function of subtask, do we want to continue with another tasks not terminate whole analysis? I think it is pointless to continue as next task is taking data from previous one. If It is a case, I think getEventRange, run() and terminate() should also use continue not returning false.

Core/JPetTaskIO/JPetTaskIO.cpp
```diff
@@ -124,6 +124,7 @@ bool JPetTaskIO::run(const JPetDataInterface&)
             return false;
           }
         }
+        fInputHandler->nextEntry();
       }
     } else {
       JPetDataInterface dummyEvent;
```
grey277: I think after line 137 we should terminate analysis if we detect error in terminate function of subtask.

Core/JPetTaskIO/JPetTaskIO.cpp
```diff
@@ -113,7 +113,7 @@ bool JPetTaskIO::run(const JPetDataInterface&)
         if (isProgressBar(fParams.getOptions())) {
           displayProgressBar(subTaskName, i, lastEvent);
         }
-        JPetData event(fInputHandler->getNextEntry());
+        JPetData event(fInputHandler->getEntry());
         ok = pTask->run(event);
         if (!ok) {
           ERROR("In run() of:" + subTaskName + ". ");
```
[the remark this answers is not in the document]
wkrzemien: ok. Added

Core/JPetTaskIO/JPetInputHandler.cpp
```diff
@@ -67,6 +56,40 @@ void JPetInputHandler::closeInput()
...
+  std::tie(isOK, firstEntry, lastEntry) = calculateEntryRange(options);
+  if (!isOK) {
+    ERROR("Some error occured in getEntryRange");
+    return false;
+  }
+  fEntryRange.firstEntry = firstEntry;
+  fEntryRange.lastEntry = lastEntry;
+  fEntryRange.currentEntry = firstEntry;
+  assert (fReader);
+  fReader->nthEntry(fEntryRange.currentEntry);
```
grey277: missing return true

Core/JPetTaskIO/JPetTaskIO.cpp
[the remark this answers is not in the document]
grey277: I'm not sure if it was intended, but I do not see any return/continue/break after this error, even thought you said it was added.

grey277:
I have small comment to line 119 of JPetTaskIO, that was marked as resolved. I'm not sure if it was intended, but I do not see any return/continue/break after this error, even thought you said it was added.

Core/JPetTaskIO/JPetTaskIO.cpp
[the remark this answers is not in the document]
wkrzemien: No, I have ovewritten it. Thanks. I am adding the fix now

Core/JPetTaskIO/JPetInputHandler.cpp
wkrzemien: Added

wkrzemien:
Ok. I think it is ok now @grey277

PR #201, opened 2018-12-17
Fix bad auto deduction variable type in JPetUnpacker

grey277:
Change auto to std::string to force std::string type after calling JPetCommonTools::stripFileNameSuffix.
Automatic variable type deduced const char *, but it was passed to std::string& and in the log you could see, that unzipped name was not correct:

72: [JPetUnpacker.cpp:exec@43] ThreadID: 0x00007fdaa221ab00 <error> The hld file doesnt exist: \C0\861\B7\ECU

Also add flag to return status of unpacking to exit program if unpacking was not successful.

Beside of that, I removed inheritance of TObject from JPetUnpacker, it produced warning:
Warning in <JPetUnpacker>: The data members of JPetUnpacker will not be stored, because it inherits from TObject but does not have its own ClassDef.
and I couldn't find anywhere where something from TObject was used in JPetUnpacker

### DIFF
--- a/Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
+++ b/Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
@@ -62,28 +62,37 @@ bool JPetUnzipAndUnpackTask::run(const JPetDataInterface&)
   auto inputFile = getInputFile(fOptions);
   auto inputFileType = FileTypeChecker::getInputFileType(fOptions);
   auto unpackerConfigFile = getUnpackerConfigFile(fOptions);
+  bool unpackFlag = false;
 
   if (inputFileType == FileTypeChecker::kHld) {
-    unpackFile(inputFile, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+    unpackFlag = unpackFile(inputFile, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+    if (!unpackFlag) {
+      ERROR("Error occurred while unpacking file: " + inputFile);
+    }
     fUnpackHappened = true;
   } else if ( inputFileType == FileTypeChecker::kZip) {
     INFO( std::string("Unzipping file before unpacking") );
     if ( !unzipFile(inputFile) ) {
-      ERROR( std::string("Problem with unpacking file: ") + inputFile );
+      ERROR( std::string("Problem with Unzipping file: ") + inputFile );
       return false;
     } else {
       INFO( std::string("Unpacking") );
-      auto unzippedFilename = JPetCommonTools::stripFileNameSuffix(std::string(inputFile)).c_str();
-      unpackFile(unzippedFilename, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+      std::string unzippedFilename = JPetCommonTools::stripFileNameSuffix(std::string(inputFile)).c_str();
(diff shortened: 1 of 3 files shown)


Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
```diff
@@ -62,28 +62,37 @@ bool JPetUnzipAndUnpackTask::run(const JPetDataInterface&)
   auto inputFile = getInputFile(fOptions);
   auto inputFileType = FileTypeChecker::getInputFileType(fOptions);
   auto unpackerConfigFile = getUnpackerConfigFile(fOptions);
+  bool unpackFlag = false;
 
   if (inputFileType == FileTypeChecker::kHld) {
-    unpackFile(inputFile, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+    unpackFlag = unpackFile(inputFile, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
```
wkrzemien: I don't like the fact that there are unpackFlag and fUnpackHappened
What is the relation between two?
Maybe one can reduce it to one variable?

Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
```diff
@@ -62,28 +62,37 @@ bool JPetUnzipAndUnpackTask::run(const JPetDataInterface&)
...
       return false;
     } else {
       INFO( std::string("Unpacking") );
-      auto unzippedFilename = JPetCommonTools::stripFileNameSuffix(std::string(inputFile)).c_str();
-      unpackFile(unzippedFilename, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+      std::string unzippedFilename = JPetCommonTools::stripFileNameSuffix(std::string(inputFile)).c_str();
+      unpackFlag = unpackFile(unzippedFilename, getTotalEvents(fOptions), unpackerConfigFile, fTOToffsetCalibFile, fTDCnonlinearityCalibFile);
+      if (!unpackFlag) {
+        ERROR("Error occurred while unpacking file: " + unzippedFilename);
+      }
```
wkrzemien: here there is clearly some inconsitenci,
It can happen thatn unpackFlag is false, error occurs and then fUnpackHapped is set to true.
Or they mean something different

Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
```diff
@@ -62,28 +62,37 @@ bool JPetUnzipAndUnpackTask::run(const JPetDataInterface&)
...
       fUnpackHappened = true;
     }
+  } else { // for any other type just set unpack flag to true
+    unpackFlag = true;
   }
 
   if (FileTypeChecker::getInputFileType(fOptions) == FileTypeChecker::kUndefinedFileType) {
     ERROR( Form("Unknown file type provided for file: %s", getInputFile(fOptions).c_str()) );
     return false;
```
wkrzemien: consequently return variable everywhere or true/ false

Tasks/JPetUnzipAndUnpackTask/JPetUnzipAndUnpackTask.cpp
wkrzemien: Since fUnpackHappend seems to be only set and not used, mabye remove it
and just assign the result of the function call 
fUnpackHappend = run()  ?

grey277:
I refactored full run method, I think now It is much more clear. I left fUnpackHapped variable, because it is used in terminate to change some userParams and added new local variable to run. 
Local variable runStatus tells if there was some error during unzipping/unpacking, but it is also true, when input file was root or other type that do not require any unzipping/unpacking. 

If input file type will be undefined I'm returning false and terminating analysis.

wkrzemien:
Very nice!!!

PR #215, opened 2019-03-15
Geant fix

daria137:
Fixed hot issues with Parser:
- added connection between reconstructed and generated Hits
- in GeantParser events are filling timewindow according to poisson distribution


daria137:
Closed this one; I will add one with reference MCrec -> MCgen (now is created otherwise)

PR #223, opened 2019-05-25
Update manual to version 7.0 - master

kkacprzak:
I have noticed, that we did not update the manual for version 7 release.


wkrzemien:
I think that adding directly to master  is going to make some problems with the history. It should be added both to master and develop. @alekgajos do you know what is the best way to do it?

kkacprzak:
Wojtek suggested adding a patch, I will try to do this.

kkacprzak:
ok I do not know how to do it, making a patch and then making a PR has the same effect as committing, pushing and making a PR. Call the gitmaster @alekgajos !

PR #238, opened 2019-11-22
JPetStatistic update

kdulski:
- THxF -> THxD (x = 1, 2, 3) requested change: Every file that uses histograms the old way with THxF needs to be adapted to ThxD
- TH3D histograms handling
- Defining histogram with axis names, not returning object to user to stop crashes. Returning only Errors in terminal and in log to notify user about errors
- Filling histograms not by returning histogram to the user, also to stop crashes, only notifications are written
- Square histograms defined as a square TCanvas. Histogram is filled separately, square histogram is drawn on the TCanvas with the same name as histogram, with addition "_Square". There is default option to draw 2D histos in colz style, because after saving there were some problems with using drawing options in TBrowser fro TCanvas


alekgajos:
It seems that your PR contains changes related to JPetLogger which are already on develop (likely due to the merge commit in the history of this PR). Please try to apply your commits on top of a branch initially consistent with develop. You can do it e.g. by fetching develop from this (upstream) repository, making a new branch from it and using `git cherry-pick` to move your three commits to this new branch, so that there are only your changes on top of the current develop.
Otherwise we cannot merge this PR without making a mess in the Git history.

kdulski:
I made a mess in a history for my master branch, so I close this PR, and open a new from clean branch. Sorry for inconvinience.

PR #240, opened 2019-12-23
Change version to c++14 in CMakeLists

wkrzemien:
Enable c++14 flag in cmake configuration.


wkrzemien:
@alekgajos I removed the comment as suggested

PR #241, opened 2019-12-30
Add seed option in JPetGeantParser to be used in gRandom

wkrzemien:
The gRandom is used both for generating lifetimes and for smearing
resolutions.
As it was found by Nikodem the default seed has fixed value,
therefore in case of e.g. multiple file processing, the same sequences
of pseudorandom values would be used.
This PR introduces: 1. an option, that the user can provide in the json file
to decide about the seed value. 2.default seed value is set to 0, so the
random generator will use the seed generated from the current system
time.
The seed values are passed to ROOT gRandom.

### DIFF
--- a/src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
+++ b/src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
@@ -72,6 +72,7 @@ bool JPetGeantParser::init()
     fSeed = getOptionAsInt(fParams.getOptions(), kSeedParamKey);
   }
   JPetGeantParserTools::setSeedTogRandom(fSeed);
+  INFO("Seed value used for resolution smearing of MC simulation data:"<< boost::lexical_cast<std::string>(fSeed));
 
   if (fMakeHisto)
     bookBasicHistograms();
--- a/src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
+++ b/src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
@@ -71,8 +71,8 @@ bool JPetGeantParser::init()
   if (isOptionSet(fParams.getOptions(), kSeedParamKey)) {
     fSeed = getOptionAsInt(fParams.getOptions(), kSeedParamKey);
   }
-  JPetGeantParserTools::setSeedTogRandom(fSeed);
-  INFO("Seed value used for resolution smearing of MC simulation data:"<< boost::lexical_cast<std::string>(fSeed));
+  JPetGeantParserTools::setSeedTogRandom(getOriginalSeed());
+  INFO("Seed value used for resolution smearing of MC simulation data:"<< boost::lexical_cast<std::string>(getOriginalSeed()));
 
   if (fMakeHisto)
     bookBasicHistograms();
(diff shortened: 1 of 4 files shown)


alekgajos:
Please add two things:
* logging of the GeantParser seed
* PID-dependent seeding in case no user-defined seed is provided

wkrzemien:
Ad1. I added login seed to file.
Ad2 According to the user guide , tuuid, which is used for calculating seed should be unique anyway:
"A TProcessID uniquely identifies a ROOT job. The TProcessID title consists of a TUUID object, which
provides a globally unique identifier. The TUUID class implements the UUID (Universally Unique Identifier), also known as GUID (Globally Unique Identifier). A UUID is 128 bits long, and if generated according to this algorithm, is either guaranteed to be different from all other UUID generated until 3400 A.D. or extremely likely to be different. ". I think it is safe to use it.

wkrzemien:
@alekgajos @Alvarness  do you think it is ok now?

Alvarness:
Setting seed seems OK. However I have two concerns:
1) As we discussed yesterday, getSeed and getSeedFromgRandom due to a fact how TRandom works
2) Both of the functions aren't called anywhere and information about seed is saved only in a log file (which I think is sufficient, but it makes those functions obsolete). Also I don't think they are user callable since GeantParser is working "under the hood" and not an additional task so I'm not sure about their purpose

src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
```diff
@@ -68,6 +68,12 @@ bool JPetGeantParser::init()
   {
     fProcessSingleEventinWindow = getOptionAsBool(fParams.getOptions(), kProcessSingleEventinWindowParamKey);
   }
+  if (isOptionSet(fParams.getOptions(), kSeedParamKey)) {
+    fSeed = getOptionAsInt(fParams.getOptions(), kSeedParamKey);
+  }
+  JPetGeantParserTools::setSeedTogRandom(fSeed);
+  INFO("Seed value used for resolution smearing of MC simulation data:"<< boost::lexical_cast<std::string>(fSeed));
```
alekgajos: I agree with Nikodem that for the `getSeed()` method to make sense you should at least use it here rather than accessing fSeed directly.

alekgajos:
Also, as Nikodem mentioned, the implementation of seed storage in `TRandom3`, which is the default gRandom really makes `TRandom::GetSeed()` reliable only before the first ever call to TRandom::Rndm() (have a look into the `TRandom3` source for how seed is stored as `fMt[0]` and how it is overwritten in Rndm...), therefore the `getSeedFromgRandom()` can never be really reliable and perhaps it's better not to add it.

src/GeantParser/JPetGeantParser/JPetGeantParser.cpp
wkrzemien: Actually his point was more about the impementation in ROOT, does not really return first seed and depends on generator used

alekgajos: I guess thi was about `getSeedFromgRandom()` rather than `getSeed()`?

wkrzemien: Ok I removed it

wkrzemien: aaa my bad. sure

wkrzemien: I removed the getSeedFromgRanodm

PR #242, opened 2020-01-10
JPetStatistic update

kdulski:
THxD histograms handling added
    TH3D histograms handling
    Defining histogram with axis names, not returning object to user to stop crashes. Returning only Errors in terminal and in log to notify user about errors
    Filling histograms not by returning histogram to the user, also to stop crashes, only notifications are written
    Square histograms defined as a square TCanvas. Histogram is filled separately, square histogram is drawn on the TCanvas with the same name as histogram, with addition "_Square". There is default option to draw 2D histos in colz style, because after saving there were some problems with using drawing options in TBrowser fro TCanvas

### DIFF
--- a/include/Core/JPetStatistics/JPetStatistics.h
+++ b/include/Core/JPetStatistics/JPetStatistics.h
@@ -20,12 +20,15 @@
 #include <THashTable.h>
 #include <TH1F.h>
 #include <TH2F.h>
+#include <TH3F.h>
 #include <TString.h>
 #include <TCanvas.h>
 #include <TGraph.h>
+#include <TClass.h>
 #include <TEfficiency.h>
 #include <string>
 #include <map>
+#include <set>
 
 /**
  * @brief Cointainer class for processing statistics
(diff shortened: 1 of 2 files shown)


src/Core/JPetStatistics/JPetStatistics.cpp
```diff
@@ -47,3 +110,14 @@ void JPetStatistics::createCounter(const char* name) { fCounters[name] = 0.0; }
 double& JPetStatistics::getCounter(const char* name) { return fCounters[name]; }
 
 const THashTable* JPetStatistics::getStatsTable() const { return &fStats; }
+
+void JPetStatistics::writeError(const char* nameOfHistogram, const char* messageEnd )
```
alekgajos: I think that grouping of identical errors is already done by JPetLogger in version 8.1, so using and checking the set here should not be necessary? Could you try if the repeated errors are grouped  also if `writeError` only calls `ERROR` from line 120, without any other logic?

In any case, please do not use cout. We have a convention that all messages should go only to the log.

include/Core/JPetStatistics/JPetStatistics.h
```diff
@@ -20,36 +20,55 @@
...
 #include <TString.h>
 #include <TCanvas.h>
 #include <TGraph.h>
+#include <TClass.h>
 #include <TEfficiency.h>
 #include <string>
 #include <map>
+#include <set>
 
 /**
```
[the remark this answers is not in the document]
kdulski: Done

src/Core/JPetStatistics/JPetStatistics.cpp
kdulski: I understand, I have write this part before Logger update, writeError reduced only to calling Error

alekgajos:
@wkrzemien I approve of this PR, please have a final look and merge if it's ok.

PR #246, opened 2020-04-28
Adding method to set custom Axes for user

kdulski:
During the change of the examples to incorporate new methods for histogram creation I`ve noticed, that some of the histograms are created with custom names of the bins (if the signal/hit are corrupted or not, or filling histogram with number of lead/trail signals at each threshold).
Therefore I propose to add additional function to JPetStatistic in order to cover these cases with one function. I tried to made it more gentle but this version of PR is the shortest one, that I can think of.

### DIFF
--- a/include/Core/JPetStatistics/JPetStatistics.h
+++ b/include/Core/JPetStatistics/JPetStatistics.h
@@ -52,9 +52,9 @@ public:
     
   enum AxisLabel
   {
-    xAxis,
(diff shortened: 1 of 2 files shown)


kdulski:
I agree I tried to use your suggestions and create function to set custom axis without creation, with usage of AxisLabel structure. I think that this approach is easier to handle. 
Let me know what you think about changes.

PR #257, opened 2020-07-05
Add root6.20 and ubuntu 18.04 to CI actions

wkrzemien:
It is just TEMPORARY version. It is artificially long because
ROOT directory name are not the same in the prepared binaries.
It should be changed.


wkrzemien:
@alekgajos would it be possible to change the name of the folder to `root` in the binary package for ROOT6.20?

alekgajos:
> @alekgajos would it be possible to change the name of the folder to `root` in the binary package for ROOT6.20?

Done.

wkrzemien:
@alekgajos ready to be merged

PR #260, opened 2020-09-17
Read json setups generated by j-pet-geant4

kkacprzak:
Json files with detector setups for MCGeantAnalysis do not have to include low level structures, as channels or photomultipliers, because they are not the part of simulations. They are set to "expected missing" in ParamManager for mcGeant file type. 
Also updating the test (new test file is on the server).


wkrzemien (approved):
It seems ok for me

wkrzemien:
@alekgajos Is it ok for you?

alekgajos:
It is ok for me as well.

PR #275, opened 2021-02-15
New Hit Class schema

kkacprzak:
This PR adds several classes, according to discussed approach for developing data reconstruction.


include/MC/JPetBaseMCHit/JPetBaseMCHit.h
```diff
@@ -0,0 +1,64 @@
+/**
```
alekgajos, line 1: I am not sure if including "base" in the class name is a good idea since it is no longer a base class. Additionally, we should somehow indicate that this is a class intended for the "True" MC.

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
```diff
@@ -0,0 +1,64 @@
...
+  TVector3 fPolarization;
+  TVector3 fMomentum;
+#ifndef __CINT__
+  int fMCDecayTreeIndex = 0;
+  int fMCVtxIndex = 0;
+  MultiplicityTag fMultiplicityTag = JPetBaseMCHit::kUnknown;
+#else
+  int fMCDecayTreeIndex;
+  int fMCVtxIndex;
+  MultiplicityTag fMultiplicityTag;
```
alekgajos, line 58: The "genGammaMultiplicity" parameter from J-PET-Geant4 can take a large spectrum of values in case the hits come from scattered photons (11, 12, 13, 21, 23, 23, ..., 101, 102, 103, 201, ...). I think an enum will not work in case of this parameter.

However, I am all for changing the name from "genGammaMultiplicity" so something shorter and more handy as we are using this parameter a lot when discussing MC simulatiions:-)

alekgajos (commented):
Good job! I just have two doubts to discuss indicated in the comments.

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
kkacprzak: I was considering "GenMC" as "generated", but in general all MC is generated by definition. So currently I have no other ideas, how to avoid this "TrueMC" slang

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
kkacprzak: Ok, I will reverse that. I suggest just "Tag" for the new name. "Gamma multiplicity" requires a long explanation anyway, so we can change it to something generic, accompanied with proper documentation.

include/DataObjects/JPetBaseHit/JPetBaseHit.h
```diff
@@ -0,0 +1,66 @@
...
+  void setPosY(double y);
+  void setPosZ(double z);
+  void setPos(double x, double y, double z);
+  void setPos(TVector3& position);
+  void setScin(JPetScin& scin);
+  void Clear(Option_t* opt = "");
+  static JPetBaseHit& getDummyResult();
+
+protected:
+#ifndef __CINT__
```
wkrzemien, line 54: Do we need this __CINT__ flags?

wkrzemien: I mean i don't know how many people really will use the old ROOT?

include/DataObjects/JPetBaseHit/JPetBaseHit.h
```diff
@@ -0,0 +1,66 @@
...
+  double fTime = 0.0;
+  double fEnergy = 0.0;
+#else
+  double fTime;
+  double fEnergy;
+#endif
+  TVector3 fPos;
+  TRef fScin = NULL;
```
wkrzemien, line 62: nullptr; ?

include/DataObjects/JPetBaseHit/JPetBaseHit.h
```diff
@@ -0,0 +1,66 @@
...
+  const JPetScin& getScin() const;
+  void setTime(double time);
+  void setEnergy(double energy);
+  void setPosX(double x);
+  void setPosY(double y);
+  void setPosZ(double z);
+  void setPos(double x, double y, double z);
+  void setPos(TVector3& position);
+  void setScin(JPetScin& scin);
+  void Clear(Option_t* opt = "");
```
wkrzemien, line 50: Is it overridden from  TObject.  Is it automatically called somewhere where we use objects? 
If it is the case, then I would change it to `void Clear(Option_t* ) override;` Without the argument, cause we are not using it anyway. If not, I would change it to `void clear(Option_t* ) override;` but here I don't insist, cause I understand that it already exists like that in `JPetHit`.

include/DataObjects/JPetMCRecoHit/JPetMCRecoHit.h
```diff
@@ -0,0 +1,42 @@
...
+ * @brief JPetMCRecoHit
+ */
+class JPetMCRecoHit : public JPetRecoHit
+{
+public:
+  JPetMCRecoHit();
+  virtual ~JPetMCRecoHit();
+  unsigned int getMCindex() const;
+  void setMCindex(unsigned int i);
+  void Clear(Option_t* opt = "");
```
wkrzemien, line 33: See above `void Clear(Option_t* ) override;`

src/DataObjects/JPetRecoHit/JPetRecoHit.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * Get the reconstruction flag
+ */
+JPetRecoHit::RecoFlag JPetRecoHit::getRecoFlag() const { return fFlag; }
+
+/**
+ * Set the reconstruction flag with enum
+ */
+void JPetRecoHit::setRecoFlag(JPetRecoHit::RecoFlag flag) { fFlag = flag; }
+
+void JPetRecoHit::Clear(Option_t*) { fFlag = JPetRecoHit::Unknown; }
```
wkrzemien, line 36: Shouldn't we also called the parent Clear method ? e.g.  JPetBaseHit::Clear(); otherwise the orginal fields are not clearead. This probably should be checked accordingly for other cases

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
wkrzemien: I think either GenMC or TrueMC is better than Base.

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
wkrzemien: Wouldn't be better to call it "HitTag" or "HitType" etc . I don't think that multiplicity is informative at all.

src/MC/JPetBaseMCHit/JPetBaseMCHit.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+int JPetBaseMCHit::getMCDecayTreeIndex() const { return fMCDecayTreeIndex; }
+
+int JPetBaseMCHit::getMCVtxIndex() const { return fMCVtxIndex; }
+
+const TVector3& JPetBaseMCHit::getPolarization() const { return fPolarization; }
+
+const TVector3& JPetBaseMCHit::getMomentum() const { return fMomentum; }
+
+JPetBaseMCHit::MultiplicityTag JPetBaseMCHit::getMultiplicityTag() const { return fMultiplicityTag; }
```
wkrzemien, line 31: Wouldn't be nicer to called it `getHitType()` or `getHitTag()`

include/MC/JPetBaseMCHit/JPetBaseMCHit.h
```diff
@@ -0,0 +1,64 @@
...
+  {
+    kPrompt,
+    k2gBack2Back,
+    k3gOPS,
+    kUnknown
+  };
+
+  JPetBaseMCHit();
+  int getMCDecayTreeIndex() const;
+  int getMCVtxIndex() const;
```
wkrzemien, line 38: Since we are changing it, could you add some comments what given function returns e.g what MCVtxIndex means. One phrase is enough and it would help people to understand it.

wkrzemien (changes requested):
It looks very nice.
What I would still add (although I don't know exactly in which class) is the paragraph describing the general inheritance tree of hits together with some rationale behind it. It would be very helpful in the future.

include/MC/JPetRawMCHit/JPetRawMCHit.h
```diff
@@ -0,0 +1,58 @@
...
+  void Clear(Option_t*) override;
+
+private:
+  TVector3 fPolarization;
+  TVector3 fMomentum;
+  int fMCDecayTreeIndex = 0;
+  int fMCVtxIndex = 0;
+  /**
+   * Gamma tag is a number with encoded information about generated photon,
+   * the convention is described in the manual of j-pet-genat4 project
```
wkrzemien, line 49: j-pet-geant4

wkrzemien (approved):
Nice work! I found one typo in the comment that might be fixed.

PR #174, opened 2018-06-15
Gate converter

lraczynski:
This is a GATE converter that converts GATE output to JPetFramework to JPetMChit class. Geometry is created based on json file.

PR #276, opened 2021-02-22
Event type fix

wkrzemien:
Our new beta-tester (@nenprio ) discovered that there is a conflict between and enum definition in JPetEventType.h and one of the newer  version of the ROOT 6.20.04). More specifically in TStructNode.h there is an enum that has a value of kUnknown as in our JPetEventType. I could have changed the name, but wanted to be more clever and I changed the definition to scoped enum.
Then,  bitwise operations didn't work anymore, so I added the overloaded operators. I hope that is all.... .
Briefly:
-replaced enum by scoped enum
-add overloaded version of & | and ~ operators
-fix tests and the code

### DIFF
--- a/include/DataObjects/JPetEventType/JPetEventType.h
+++ b/include/DataObjects/JPetEventType/JPetEventType.h
@@ -25,4 +25,9 @@ enum class JPetEventType
   kScattered = 16,
   kCosmic = 32
 };
+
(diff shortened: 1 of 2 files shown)


alekgajos (approved):
I think all of these changes are fine.

kkacprzak (approved):
I don't have any comments to those changes. 
But this requires small update in examples repository (adding namespaces in some files).

wkrzemien:
@kkacprzak good catch! I will prepare the PR for examples as well

wkrzemien:
@kkacprzak  https://github.com/JPETTomography/j-pet-framework-examples/pull/174

PR #279, opened 2022-05-11
Fix GH Actions CI tests

alekgajos:
This PR replaces the Ubuntu-16 image (no longer supported) with Ubuntu-20 and adds a new test based on Ubuntu-22.
Additionally, it fixes a missing explicit include in JPetWriterTest.cpp which is required with new ROOT.

### DIFF
--- a/tests/Core/JPetWriter/JPetWriterTest.cpp
+++ b/tests/Core/JPetWriter/JPetWriterTest.cpp
@@ -29,6 +29,7 @@
 
 #include <TFile.h>
 #include <TList.h>
+#include <THashTable.h>
 #include <TNamed.h>
 #include <iostream>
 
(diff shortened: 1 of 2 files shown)


tests/Core/JPetWriter/JPetWriterTest.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include <TFile.h>
 #include <TList.h>
+#include <THashTable.h>
```
wkrzemien, line 32: This include  is needed? Or is it an accident?

wkrzemien: I have just seen the commit comment

alekgajos: With a new ROOT version, 6.26/02 or due to strictness of the new g++ (11.2), one of the tests failed to compile as THashTable was used but not included. Apparently, previously it was still there from a transitive include but with either the new ROOT or new compiler it's no longer the case. The problem was revealed by the new test based on Ubuntu 22.04.

wkrzemien (approved):

PR #280, opened 2022-05-13
[WIP] Fix j-pet event

wkrzemien:

PR #287, opened 2022-11-30
Fix lifetimes in JPetEvent

alekgajos:
This is the PR #280 by @wkrzemien, rebased onto the `modular4` branch and with minor conflicts resolved.
The original PR solves the problem with lifetime of hit objects contain within a JPetEvent which was introduced after moving from storing a vector of hit objects to a vector of pointers.

### DIFF
--- a/include/DataObjects/JPetEvent/JPetEvent.h
+++ b/include/DataObjects/JPetEvent/JPetEvent.h
@@ -53,7 +53,7 @@ public:
 
   JPetEvent();
   JPetEvent(const std::vector<const JPetBaseHit*>& hits, JPetEventType eventType = JPetEventType::kUnknown, bool orderedByTime = true);
-  ///@TODO add move and = operators and virtual disestructor (rule of five!)
+  ~JPetEvent()=default; // despite the presence of assignment and move operators, there are no resources to free manually
   JPetEvent(const JPetEvent& other);
   JPetEvent& operator=(const JPetEvent& other);
   JPetEvent(JPetEvent&& other);
--- a/include/DataObjects/JPetEvent/JPetEvent.h
+++ b/include/DataObjects/JPetEvent/JPetEvent.h
@@ -56,6 +56,8 @@ public:
   ///@TODO add move and = operators and virtual disestructor (rule of five!)
   JPetEvent(const JPetEvent& other);
   JPetEvent& operator=(const JPetEvent& other);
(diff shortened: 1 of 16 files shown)


include/DataObjects/JPetEvent/JPetEvent.h
```diff
@@ -53,8 +53,13 @@ class JPetEvent : public TObject
 
   JPetEvent();
   JPetEvent(const std::vector<const JPetBaseHit*>& hits, JPetEventType eventType = JPetEventType::kUnknown, bool orderedByTime = true);
+  ///@TODO add move and = operators and virtual disestructor (rule of five!)
```
alekgajos, line 56: @wkrzemien, I have added a move constructor and assignment operator. However, despite the rule of 5, I don't think a custom destructor is needed here as all heap memory is managed with unique pointers.
Please let me know what you think about it.

src/DataObjects/JPetEvent/JPetEvent.cpp
```diff
@@ -25,35 +28,86 @@ JPetEvent::JPetEvent(const std::vector<const JPetBaseHit*>& hits, JPetEventType
   setHits(hits, orderedByTime);
 }
 
+JPetEvent::JPetEvent(const JPetEvent& other)
+{
+  if (&other != this)
```
alekgajos, line 33: @wkrzemien, It seems to me that in case of constructors (the same is used in assignment operators and is justified there) this check is redundant as "other" should never be the same as "this". Could you confirm it?

include/DataObjects/JPetEvent/JPetEvent.h
wkrzemien: @alekgajos You are right. Maybe you could add something like:
`~JPetEvent()=default;` with a comment that even if the class has defined move or assignment operators we don't need to release any resources now.

src/DataObjects/JPetEvent/JPetEvent.cpp
wkrzemien: Yes, you are right. It should be removed.

alekgajos: Removed.

include/DataObjects/JPetEvent/JPetEvent.h
alekgajos: Done.

